PR #428, opened 2013-03-20
Implement primitive-machine-word-count-ones

waywardmonkeys:
HARP backend support isn't in place yet.

This isn't ready for actual merging ... but it should be ready for review if anyone is interested, apart from the HARP backend support.


waywardmonkeys:
I've updated this against current master.

waywardmonkeys:
@housel: I have finally implemented this for HARP. I have not built or run it with the HARP back-end yet.

This is ready for review and if someone can run it on 32 bit with HARP and see how it goes, that would be great. I'd like to see this in 2015.1 as it is useful for implementing some other fun stuff.

waywardmonkeys:
This has been updated to use this new primitive in the implementation of ``bit-count`` for ``<bit-vector>``.

housel:
Manually merged as commit 6f60c211.

PR #647, opened 2014-01-18
At least on my FreeBSD the struct timeval consists of two longs

hannesm:
(thus 64bit on my amd64 system).

Any operating system which does this differently? Linux? MacOSX?


waywardmonkeys:
I have a new network library that will replace this one. It doesn't have this issue as it uses melange to generate per-system definitions.

PR #806, opened 2014-08-08
REVIEW: NOT FOR MERGE: Function types

waywardmonkeys:
This is NOT FOR MERGE. This is just to give a way to keep track of where things are at with this work.


waywardmonkeys:
Going to submit a new PR with these changes in an updated cleaner form.

PR #808, opened 2014-08-13
Multi line strings

cgay:
FOR REVIEW ONLY

If you have time to take a closer look and comment that would be great.  I added a couple more assertions and fixed a couple of bugs related to newlines.  Source locations work now.  (That is, if there's an error _following_ a multi-line string, it is no longer off by the number of newlines in the string.)

I still need to canonicalize end-of-line sequences...


pedro-w:
I checked this out, and, in my limited experience, it looks good. I couldn't make it go wrong by giving it  incorrect inputs.

PR #838, opened 2014-11-30
dfmc-optimization: Warn when inline-only inlining fails

housel:
This is provided in case somebody wants to play with it, but it should probably not be merged until we have solutions for the many additional warnings this adds to the build.


waywardmonkeys:
If the implementation of ``inline-only`` were better, it would inline in more cases than it does now and probably reduce the number of warnings. :)

PR #862, opened 2015-02-02
[dfmc] Remove *retract-dfm?* variable.

waywardmonkeys:
This is always true and there's nothing that flips it to #f.

### DIFF
--- a/sources/dfmc/back-end/back-end-library.dylan
+++ b/sources/dfmc/back-end/back-end-library.dylan
@@ -107,7 +107,6 @@ define module dfmc-back-end
     heap-stats,
     single-heap-stats,
     diff-heap-stats,
-    *retract-dfm?*,
 
     library-imported-object?,
     library-imported-binding?,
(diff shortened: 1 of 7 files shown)


housel:
I think it's meant to be manually flipped to #f by someone using the debugger.

waywardmonkeys:
For what purpose? I figured it was an old debugging construct from The Ancient Days.

If we go back to The Ancient Days, we can find it used in ``Sources/dfmc/file-compiler/debug.dylan``:

````dylan
define method compile-source-record (name :: <string>, library-key :: <library-designator>, #rest flags)
  let retract? = *retract-dfm?*;
  block()
    *retract-dfm?* := #f;
    close-project(library-key);
    compile-library(library-key, force-compile?: #t, skip-emit?: #t, skip-link?: #t, strip?: #f);
    apply(emit-source-record, name, library-key, flags);
    apply(link-source-record, name, library-key, flags);
  cleanup
    *retract-dfm?* := retract?;
  end block;
end method;
````

But all of that code is long long gone and dead.

housel:
It wouldn't be visible in the code; flipping this would just allow scrounging through DFM using the inspector even after code generation for that compilation record is complete.

PR #864, opened 2015-02-14
Build script and C code support for the LLVM run-time

housel:


sources/lib/run-time/exceptions.c
```diff
@@ -53,6 +53,7 @@ MMError dylan_init_thread(void **rReturn, void *(*f)(void *, size_t), void *p, s
 }
 
 
+#ifndef OPEN_DYLAN_BACKEND_LLVM
 void *dylan_callin_handler(void *arg_base, size_t s)
```
waywardmonkeys: This could be ``#ifdef OPEN_DYLAN_BACKEND_HARP`` ...

sources/lib/run-time/llvm-posix-threads.c
```diff
@@ -0,0 +1,949 @@
...
+  pthread_cond_broadcast(&thread_exit_event);
+  pthread_mutex_unlock(&thread_join_lock);
+
+  return result;
+}
+
+// primitive-make-thread
+dylan_value primitive_make_thread(dylan_value t, dylan_value n, dylan_value p, dylan_value f, DBOOL s)
+{
+  struct KLthreadGYthreadsVdylan *thread
```
waywardmonkeys: This is missing the code that we have in the C run-time for setting the thread name.

It is also missing setting the priority, but we don't do that in the C run-time either. Maybe both could just be marked as TODOs?

waywardmonkeys: Come to think of it, I think we set the thread name elsewhere ... I'd have to look it up.

housel: I'll add TODOs.

waywardmonkeys:
I'll merge this once the *.opendylan.org machines are back online.

PR #870, opened 2015-03-13
Trap arithmetic exceptions on Mac OS X

housel:


sources/lib/run-time/llvm-exceptions.c
```diff
@@ -148,14 +148,185 @@ void RemoveDylanExceptionHandlers(void)
...
+    abort();
+  }
+
+  primitive_reset_float_environment();
 }
 
 void RemoveDylanExceptionHandlers(void)
 {
-  // FIXME
+  // Nothing to do (thread is about to exit)
```
waywardmonkeys: Is this the case for a "call-in" to the Dylan world from foreign code?

housel: That depends... calling into Dylan from the main thread, or from a thread started using make(<thread>), should be just fine. Calling into Dylan from a foreign thread would require providing a trampoline interface which would take care of this sort of thing as well as cooperation with the GC. If there is a need for such an interface we can reexamine what this function does.

PR #873, opened 2015-03-24
[dfmc-conditions] Make <program-error> a <program-note>.

waywardmonkeys:
There has been a long-standing TODO item in the codebase (since
April of 1998) to make <program-error> inherit from ``<program-note>``
instead of ``<serious-program-warning>``.

* sources/dfmc/conditions/hierarchy.dylan: Change superclass for
   ``<program-error>`` to ``<program-note>``.

* sources/dfmc/conditions/presentation.dylan
  (do-with-program-conditions): Update the condition handler to
   handle all ``<program-note>`` instances rather than only
   ``<program-warning>`` instances. This allows it to handle
   ``<program-error>`` instances as well as (in the future) other
   sorts of ``<program-note>``.

* sources/dfmc/debug-back-end/print-condition
  (condition-classification): Return the correct text for
   ``<program-error>`` instances.

* documentation/hacker-guide/source/compiler/notes-warnings-errors.rst:
   Update superclass for ``<program-error>`` and remove note about the
   TODO item.

### DIFF
--- a/documentation/hacker-guide/source/compiler/notes-warnings-errors.rst
+++ b/documentation/hacker-guide/source/compiler/notes-warnings-errors.rst
@@ -390,18 +390,13 @@ Program Conditions
    :open:
    :abstract:
 
-   :superclasses: :class:`<serious-program-warning>`
+   :superclasses: :class:`<program-note>`
 
    :description:
 
      A ``<program-error>`` is a language error.  Examples would be (most)
      syntax errors, inconsistent direct superclasses, or a reference to
      an undefined name.
-
-     .. note:: This is currently a :class:`<serious-program-warning>`
-        rather than a separate subclass of :class:`<program-note>`.
-        This is noted in the source as a short term hack since
-        April 6, 1998.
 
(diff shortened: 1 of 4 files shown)


housel:
Can you speculate why this wouldn't have been done before?

waywardmonkeys:
I'm not sure.  Speculation:

* The old code path, as it wasn't didn't work in the native environment (just the emulator, I think). However, he made the same hierarchy change in the emulator code.
* The change was made in April, 1998 when he was making some other changes to condition handling / warnings, but nothing that seems too related to this.
* Nothing in the core compiler actually uses ``<program-error>`` ... they're only used in ``dfmc-c-ffi``.

With my changes, they show up as ``Error`` rather than ``Serious warning``, but they still use the same reporting code path as serious warnings, so they don't actually abort the compiler. That'll be a follow up change. The reason for that is that ``$signal-program-error`` doesn't actually work currently. (It doesn't crash, it just doesn't display anything or abort the compilation.)

PR #874, opened 2015-03-28
Float invalid

waywardmonkeys:
This can NOT be merged as it breaks bootstrapping due to issues with ``llvm-runtime-generator``.


waywardmonkeys:
I would like to write tests for this ... but since I can't bootstrap it, that's difficult.

Also, this was landed in similar form (minus LLVM changes) previously and reverted for an unknown reason, perhaps something was failing in MPS builds with HARP?

waywardmonkeys:
This should be good to go now.

PR #880, opened 2015-04-12
Move format-to-string: part two

waywardmonkeys:
This is a continuation of some changes that are in the middle of being made to clean up some of the ``format-to-string`` mess in ``common-dylan``.

@housel: It would be ideal if this were tested on Win32 prior to landing.


housel:
x86-win32 release build completed with no format-related serious warnings.

PR #890, opened 2015-04-25
[common-dylan] Add float classification.

waywardmonkeys:
Add a `classify-float` that returns whether the float is a NaN, infinite,
zero, subnormal or normal value.

* sources/common-dylan/library.dylan
  (module `common-extensions`): Export `classify-float`, `<float-classification>`.

* sources/common-dylan/numerics.dylan
  (``<float-classification>``): New.
  (`classify-float`): New.

* documentation/release-notes/source/2015.1.rst: Update.

### DIFF
--- a/documentation/release-notes/source/2015.1.rst
+++ b/documentation/release-notes/source/2015.1.rst
@@ -98,6 +98,9 @@ Common Dylan
   * ``asin``, ``acos``, ``atan``
   * ``sinh``, ``cosh``, ``tanh``
   * ``asinh``, ``acosh``, ``atanh``
+* The ``common-dylan`` library now provides a ``classify-float``
+  method which will return if the given float is ``#"normal"``,
+  ``#"zero"``, ``#"infinite"``, ``#"nan"``, or ``#"subnormal"``.
 
 Compiler
 ========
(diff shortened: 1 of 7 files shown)


waywardmonkeys:
This is mainly for review by @housel. I plan to add some tests and perhaps even some documentation ... but I don't mind if they become follow-on commits and/or pull requests.

waywardmonkeys:
I think there's probably a good conversation that could be had about where to place ``classify-float`` ... should it be just in ``common-dylan`` like I have it here? Or should it be part of the ``dylan-extensions`` module of ``dylan``?

The old plan from Harlequin days for ``classify-float`` was that it was a compiler primitive and exported from ``dylan-extensions``.  (I've removed this code in #891.)

housel:
Looks fine so far (including the placement in common-dylan). I'll wait for tests to be added before merging.

waywardmonkeys:
@housel: Want to take a look again?

Note that the ``<double-float>`` tests do NOT work on the C back-end. This is due to a bug reading large integer values elsewhere. (And I'd appreciate help fixing that...)

waywardmonkeys:
@housel Ping again. I'd like to see this land. As I indicated before, the failing test case on 64 bit C backend is a separate / known bug.

PR #894, opened 2015-05-05
[system,settings] Bootstrapping cleanup.

waywardmonkeys:
On July 30, 2001, Gary Palter added a comment for some maintenance work
that should happen once Functional Developer 2.1a1 had shipped.  It
shipped a while ago, so we can do this cleanup now.

* sources/system/library.dylan
  (module settings-internals): No longer export invalidate-settings-caches
    or %settings-key-name-setter.

* sources/system/settings/settings.dylan
  (generic invalidate-settings-caches): Mark as dynamic and not open.
  (generic %settings-key-name-setter): Mark as dynamci and not open.

### DIFF
--- a/sources/system/library.dylan
+++ b/sources/system/library.dylan
@@ -332,11 +332,6 @@ define module settings-internals
          register-key, unregister-key,
          get-value, set-value,
          do-remove-value!;
-
-  //---*** BOOTSTRAPPING: Remove these exports and declare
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This PR has been updated for the comment from @housel.

waywardmonkeys:
I'm closing this for now. I'm no longer sure what the right next step here is.

PR #895, opened 2015-05-07
Alter how the debugger-nub is linked in on Windows.

waywardmonkeys:
Previously, we just linked against a library on Windows and had
a custom access-path LID file to pull that library in. Instead,
we can have a debugger-nub Dylan library that knows to link
against devnub.lib and will pull that in.

* sources/registry/x86-win32/access-path: Remove as there is no
    Windows specific access-path LID file any longer.

* sources/registry/x86-win32/debugger-nub: New registry entry for
    the new debugger-nub Dylan library.

* sources/runtime-manager/access-path/library.dylan
  (library access-path): Use the new debugger-nub library to pull it in.

* sources/runtime-manager/access-path/win32-access-path.lid: Remove
    as it is no longer needed since the C libraries are pulled in
    via the debugger-nub Dylan library.

* sources/runtime-manager/debugger-nub/x86-win32/library.dylan: New.

* sources/runtime-manager/debugger-nub/x86-win32/x86-win32-debugger-nub.lid: New.

### DIFF
--- a/sources/registry/x86-win32/access-path
+++ b/sources/registry/x86-win32/access-path
@@ -1,1 +1,0 @@
-abstract://dylan/runtime-manager/access-path/win32-access-path.lid
--- a/sources/registry/x86-win32/debugger-nub
+++ b/sources/registry/x86-win32/debugger-nub
@@ -1,0 +1,1 @@
+abstract://dylan/runtime-manager/debugger-nub/x86-win32/x86-win32-debugger-nub.lid
--- a/sources/runtime-manager/access-path/library.dylan
+++ b/sources/runtime-manager/access-path/library.dylan
@@ -15,6 +15,7 @@ define library access-path
   use io;
   use system;
   use c-ffi;
+  use debugger-nub;
 
   export
     access-path, access-path-nub;
--- a/sources/runtime-manager/access-path/win32-access-path.lid
+++ b/sources/runtime-manager/access-path/win32-access-path.lid
@@ -1,9 +1,0 @@
-library:     access-path
-LID:         access-path.lid
-C-libraries: devnub.lib
-             minvcrt.lib
-             dbghelp.LIB
-Copyright:    Original Code is Copyright (c) 1995-2004 Functional Objects, Inc.
-              All rights reserved.
-License:      See License.txt in this distribution for details.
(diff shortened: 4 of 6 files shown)


waywardmonkeys:
This lays the groundwork for subsequent changes which will be in a new pull request.

I haven't tested that this actually builds on Windows (no Windows build environment).

housel:
Could we hold off on this a bit? A few weeks ago I started a branch that takes a different approach to this, and I should have time to work on it next week. See https://github.com/housel/opendylan/tree/lldb-nub to get an idea of where it is going (though the local-access-path library isn't compilable yet).

waywardmonkeys:
I've also got an LLDB nub underway now that I can build the C code.

I like what you did in https://github.com/housel/opendylan/commit/54f44b14e6c518c40fc733bb1a686e38184daf95 and would like to see that move to master (and then I'd update the skeleton and lldb-nubs).

I'll have to take a deeper look at what is happening in https://github.com/housel/opendylan/commit/5406b4d9baccd5a983792e151dc922729b75bfaa

With LLDB, we don't need a distinction between a local ``access-path`` and a remote one since a single LLDB access path can (eventually) handle local and remote. (Eventually because this doesn't always work correctly in LLDB yet... early days for LLDB still.)

waywardmonkeys:
What I have for talking to LLDB is a compilable version of the nub interface that exports a C API and then uses the public (``SB*`` classes) API to LLDB. This is the same sort of approach that Xcode and others take when integrating with LLDB.

housel:
The remote-access-path library uses CORBA to talk to the debugger-server application. At this point on my branch it is possible to compile the remote-access-path library without needing to link to any C code; I can then implement the same CORBA interface in Python making use of the lldb bindings.

waywardmonkeys:
I understand that's what you're doing ... but I see value in a straight-forward interaction with LLDB that doesn't involve a lot of additional complexity so I'm taking that approach with the local access-path for now.

I'll continue on a branch for now and I'll cherry-pick https://github.com/housel/opendylan/commit/54f44b14e6c518c40fc733bb1a686e38184daf95 until it lands on master. (But that should be able to land separately from the other work!)

waywardmonkeys:
@housel has landed a different approach to this on master and this is no longer needed.

Issue #897, opened 2015-05-08
Optimization to integer addition


waywardmonkeys:
When we add 2 integers, we avoid untagging and retagging them by doing this:

````dylan
define sealed inline method \+ (x :: <integer>, y :: <integer>)
 => (result :: <integer>)
  let mx = interpret-integer-as-machine-word(x);
  let my = strip-integer-tag(interpret-integer-as-machine-word(y));
  let result = machine-word-add-signal-overflow(mx, my);
  interpret-machine-word-as-integer(result)
end method \+;
````

The definition of ``strip-integer-tag`` is:

````dylan
define inline-only function strip-integer-tag
    (x :: <machine-word>) => (result :: <machine-word>)
  machine-word-logxor(x, $integer-tag-value)
end function strip-integer-tag;
````

This results in this sort of thing in the generated assembly (x86_64):

````asm
    0x100003e6f <+47>: xorq   $0x1, %rax
    0x100003e73 <+51>: addq   %rax, %r14
````

@pkhuong has suggested that we could get better generated code if we did a subtraction instead of an xor operation here as that would allow LLVM to do better optimization:

> [19:33:08]  \<pkhuong>	brucem: I think I kind of see what's going on.
> [19:33:30]  \<pkhuong>	the xor is messing up llvm's detection of commutativity
> [19:34:17]  \<pkhuong>	it does r14 <- rax + r14; r14 -> rbx; rbx -> rax
> [19:34:49]  \<pkhuong>	ideally, you'd get rax <- rax + r14

This seems like an interesting experiment to try.

It also pointed out an issue where not all back-ends might be using tagged integers (like my JS back-end might not).

It seems like an easy way to address this would be to add a new ``primitive-strip-integer-tag`` and perhaps for some other similar operations and then each back-end can do the right thing.

waywardmonkeys:
I also took the ``.ll`` file and hand edited it to change from ``xor`` to ``sub`` and passed that to ``clang -O2 -c -S`` and then diffed the assembly outputs:

````diff
batavia:fib bruce$ diff -u fib.xor.S fib.sub.S
--- fib.xor.S	2015-05-08 20:43:43.000000000 +0700
+++ fib.sub.S	2015-05-08 20:44:09.000000000 +0700
@@ -34,7 +34,7 @@
 ## BB#3:
 	movq	%rbx, %rdi
 	callq	_KfibVfibI
-	xorq	$1, %rax
+	decq	%rax
[...]

PR #898, opened 2015-05-08
Optimize integer tag manipulation

waywardmonkeys:
This is an experiment to address what is discussed in issue #897.

### DIFF
--- a/sources/dylan/conversion-tagged-integer.dylan
+++ b/sources/dylan/conversion-tagged-integer.dylan
@@ -34,12 +34,12 @@ end function interpret-integer-as-machine-word;
 
 define inline-only function strip-integer-tag
     (x :: <machine-word>) => (result :: <machine-word>)
-  machine-word-logxor(x, $integer-tag-value)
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
(I think this is for discussion for now rather than merging right away.)

waywardmonkeys:
Some open questions:

* Should we also modify ``insert-integer-tag``?
* Does it matter that ``strip-integer-tag`` is no longer idempotent?

housel:
- Yes, I think we should do both
- I don't think it matters; "strip" should mean "strip"

Any conclusions on what effect this has on the generated C code?

waywardmonkeys:
I don't have performance numbers but the disassemblies that I put in the associated issue #897 are from the C back-end.

waywardmonkeys:
I think we can defer making this an actual compiler primitive until such a thing is actually needed.

waywardmonkeys:
This has been updated and should be ready for merging. It makes it through a 3 stage bootstrap on Mac OS X for me.

PR #902, opened 2015-05-10
Add support for building C++ files from a LID.

### LINKED ISSUE #900: Building C++ files with a LID file
Previously, I just put C++ source files into my LID's ``c-source-files`` keyword and relied upon clang looking at the extension to know what language to use.

I now have a problem where this longer works.

I need to build my C++ files with ``-std=c++11`` or later. If I add that to my ``CCFLAGS`` though, then it fails to build correctly as it is building ``_glue.c`` with those flags and clang errors when comcpiling C code with a C++ ``-std`` option.

I also can not just put ``-xc++`` into the ``CFLAGS`` as any C files that include ``run-time.h`` like ``_glue.c`` will fail to build.

The easiest option here (perhaps) is to add a new LID keyword and pass that through to JAM.  On the LID side, we could call it ``cxx-source-files:`` and on the JAM side, perhaps ``DylanLibraryCxxSources``. We'd also want ``CXXFLAGS`` to match ``CCFLAGS``

waywardmonkeys:
Fixes #900.

* documentation/hacker-guide/source/build-system.rst: Update to
    mention DylanLibraryCxxSources.

* documentation/release-notes/source/2015.1.rst: Mention new support.

* sources/environment/project-wizard/project-description.dylan
  ($project-file-header-order): Add #"cxx-source-files".

* sources/jamfiles/config.jam.in
  (CXX, CXXFLAGS): New variables.

* sources/jamfiles/mini-jambase.jam
  (rule Cxx, actions Cxx): New.

* sources/jamfiles/posix-build.jam:
  (rule DylanLibraryCxxSources): New.

* sources/jamfiles/x86-win32-pellesc-build.jam
  (CXX, CXXFLAGS): New variables.

* sources/jamfiles/x86-win32-vc6-build.jam
  (CXX, CXXFLAGS): New variables.
  (rule DylanLibraryCxxSources): New.

* sources/jamfiles/x86-win32-vc8-build.jam
  (CXX, CXXFLAGS): New variables.
  (rule DylanLibraryCxxSources): New.

* sources/lib/build-system/jam-build.dylan
  (jam-read-mkf): Add #"cxx-source-files" handling.

* sources/project-manager/projects/lid-projects.dylan
  ($standard-lid-keyword, $list-build-keyword): Add #"cxx-source-files".
  (copy-extra-records): Copy C++ source files.

* sources/project-manager/user-projects/add-file.dylan
  (project-other-sources): Include C++ sources.
  (project-add-file-of-type): Support C++ extensions.
  (project-remove-file-of-type): Support C++ extensions.

* sources/project-manager/user-projects/save-project.dylan
  (lid-project-build-settings): Return C++ source files.
  (save-lid-file-info): Save C++ source files.

* sources/project-manager/user-projects/user-projects.dylan
  ($cxx-source-file-types): New constant for C++ file extensions.

*

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -104,6 +104,7 @@ bootstrap_target=3-stage-bootstrap
 AC_SUBST(bootstrap_target)
 
 AC_PROG_CC([clang-3.5 clang gcc cc])
+AC_PROG_CXX([clang++-3.5 clang++ g++ c++])
 
 if test -n "$SUPPORT_LLVM"; then
     AC_MSG_CHECKING([whether Clang is new enough])
--- a/documentation/hacker-guide/source/build-system.rst
+++ b/documentation/hacker-guide/source/build-system.rst
@@ -229,6 +229,9 @@ component is the base name of the executable or shared library (from the
     normally used to ensure that files of various sorts (not just C header
     files) are copied into the build directory.
 
+``DylanLibraryC++Sources *image* : *sources* ;``
+    Link C++ source files into the shared library.
+
 ``DylanLibraryRCFiles *image* : *rcfiles* ;``
     Link Win32 resource files into the shared library and executable.
 
--- a/documentation/release-notes/source/2015.1.rst
+++ b/documentation/release-notes/source/2015.1.rst
@@ -37,6 +37,9 @@ Build System
   back-ends. These shouldn't be changed unless you're sure you know
   what you're doing.
 
+* A new option for building C++ code, ``c++-source-files``,  has been
+  added to LID files.  This matches the ``c-source-files:`` keyword.
+
 C-FFI
 =====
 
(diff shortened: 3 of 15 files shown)


waywardmonkeys:
This isn't quite ready to be merged yet. There's one last issue: When a ``DylanLibraryCxxSources`` is used with a library, we need to link it with CXX and not CC. (And at that point, we'd have to have CXX be ``g++`` or ``clang++``. Just using ``$(CC) -xc++`` doesn't work.

housel:
It should be enough to just add -lc++ to the link libraries.

waywardmonkeys:
I've updated this now with new naming using ``C++`` and such rather than ``CXX``.

As for linking, what I have to do currently is ``-lc++ -lc++abi`` ... but that won't be the case on Linux (``-lstdc++``) and on FreeBSD, I think it depends on the version of FreeBSD at play.

PR #906, opened 2015-05-18
[dylan] Remove a function call from dispatch code.

waywardmonkeys:
This changes a function returning a vector into just
the constant value for that vector. This avoids a method
call in a critical code path.

* sources/dfmc/modeling/namespaces.dylan
  (module dispatch-engine): %second-hash-values is
    now $second-hash-values.

* sources/dylan/new-dispatch.dylan
  (%second-hash-values): Convert to a constant instead,
    $second-hash-values.
  (%hckd-hash-step): Use the new constant instead of the
    function call.


waywardmonkeys:
There might be a problem with this, so hold off ...

waywardmonkeys:
Never mind, all is well.

PR #913, opened 2015-05-29
Stack walking

waywardmonkeys:


waywardmonkeys:
This adds a (Darwin-only for now) implementation based on ``libunwind``. No linking changes are required at the moment as the system library includes ``libunwind``.

Issue #921, opened 2015-06-11
-O0 vs -O2: -O0 results in miscompilation?


waywardmonkeys:
I was trying building ``melange`` by modifying my ``config.jam`` to tell the compilation to use ``-g -O0`` instead of ``-g -O2``.

This resulted in an error at run-time that wasn't present before:

```
24 is not of type {<class>: <application-process>}
24 is not of type {class <application-process> 0x57e39c}
```

(Getting this error requires some code that I haven't landed yet.)

waywardmonkeys:
This happened with this code:

```dylan
define function get-compiler-defines (cmd :: <string>)
  local method parse-define (line :: <string>)
          let define-start = size("#define ");
          let sep = find-substring(line, " ", start: define-start);
          let def = copy-sequence(line, start: define-start, end: sep);
          let val = copy-sequence(line, start: sep + 1);
          values(def, val)
        end;
  with-application-output (stream = cmd)
    let output = read-to-end(stream);
    let lines = split(strip(output), "\n");
    let defines = make(<stretchy-vector>);
    do(method(line)
        let (d, v) = parse-define(line);
        unless (find-substring(v, "##"))
          add!(defines, d);
          add!(defines, v);
        end;
       end, lines);
    defines
  end;
end function;
```

The ``with-application-output`` is:

```dlan
define macro with-application-output
  { with-application-output (?stream:variable = ?command:expression,
                             #rest ?keys:expression)
      ?body:body
    end }
  => { begin
         let _stream = #f;
         let _process = #f;
         block ()
           let (exit-code, signal, process, ostream)
             = run-application(?command, asynchronous?: #t, output: #"stream",
                               ?keys);
           _stream := ostream;
           _process := process;
           let ?stream :: <file-stream> = ostream;
           ?body;
         cleanup
           if (_process) wait-for-application-process(_process) end;
           if (_stream & stream-open?(_stream)) close(_stream) end;
         end
       end }
end macro with-application-output;
```

```dylan
METHOD get-compiler-include-directories (cmd :: <string>) => (#rest results)
  {{ _stream }} := make-cell(^#f)
  {{ _process }} := make-cell(^#f)
  UNWIND-PROTECT (t3)
[...]

PR #926, opened 2015-06-17
[dfmc-c-back-end] UWP needs volatile locals.

waywardmonkeys:
Previously, we only made locals volatile when the function
body contained a block with a non-local entry state. They
actually should be volatile if there is any ``<unwind-protect>``
involved as well.

* sources/dfmc/c-back-end/c-emit-object.dylan
  (emit-lambda-body-using-function): Factor out code and call
    ``need-volatile-locals?``.

* sources/dfmc/c-back-end/c-emit-c-object.dylan
  (emit-lambda-body-using-function): Factor out code and call
    ``need-volatile-locals?``.

* sources/dfmc/c-back-end/c-back-end.dylan
  (need-volatile-locals?): New function factored out from the
    above, but also checks for ``<unwind-protect>``.

### DIFF
--- a/sources/dfmc/c-back-end/c-back-end.dylan
+++ b/sources/dfmc/c-back-end/c-back-end.dylan
@@ -210,3 +210,21 @@ define method retract-local-methods-in-heap(heap) => ()
     end for;
   end if;
 end method;
+
+// In the C back-end, due to the use of setjmp / longjmp, we need to flag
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
This doesn't actually fix #921 which is when I had noticed this, but I'm pretty sure this is still the correct thing to do. It might be worth thinking the run-time scenarios through though to be sure.

PR #930, opened 2015-06-24
Abstract literal fragments

waywardmonkeys:
This passes a  3-stage bootstrap on Mac OS X.


waywardmonkeys:
FWIW, this code has been like this since Keith Playford checked it in during January, 1997.

PR #934, opened 2015-06-29
[console] Allow showing properties from command line.

waywardmonkeys:
* sources/environment/commands/properties.dylan
  (parse-next-argument on <command-property>): New method for parsing
    an argument and converting it to a <command-property>.

* sources/environment/console/command-line.dylan
  (class <basic-main-command>): Add slot for %show.
  (do-execute-command): Handle -show being passed by invoking the
    show property command.

* sources/environment/console/compiler-command-line.dylan
  (command-line main): Add new show keyword parameter.

* sources/environment/console/environment-command-line.dylan
  (command-line main): Add new show keyword parameter.


waywardmonkeys:
I was adding this as I'd like to be able to query ``dylan-compiler`` from a script to get back useful information ... like the current set of registries that would be searched, or where the install location is.

Unfortunately, this patch isn't really sufficient as properties are written for humans to read, so we perhaps need machine-readable property output. Also, error handling isn't all that great yet.

An alternative would be to just handle the set of values that I actually need here rather than making it fully general with the properties system.

PR #950, opened 2015-07-29
Coloring stream

waywardmonkeys:
This is an initial pull request in case someone wants to do some review.

I'll be updating this over the next day or three with further updates.

### DIFF
--- a/documentation/library-reference/source/coloring-stream/index.rst
+++ b/documentation/library-reference/source/coloring-stream/index.rst
@@ -1,0 +1,240 @@
+***************************
+The coloring-stream library
+***************************
+
+.. current-library:: coloring-stream
+.. current-module:: coloring-stream
+
+The ``coloring-stream`` module provides a means for writing text to
+the console or terminal with colors and varying intensity.
+
+On Unix, this is typically done via ANSI codes.  Windows support is
+yet to be implemented.
+
+Usage
+*****
+
+It is implemented as a :class:`<wrapper-stream>`, so it can wrap any
+given stream. While it can be created via :drm:`make`, it is easiest
+to just call :gf:`colorize-stream` on a stream to get back a
+:class:`<coloring-stream>` wrapper that can be used to output color.
+
+If the underlying stream doesn't support color, then the returned
+:class:`<coloring-stream>` will silently drop all text attributes.
+
+Once you have a :class:`<coloring-stream>`, you can set text attributes
+by writing instances of :class:`<text-attributes>`. Text attributes
+are created with :gf:`text-attributes`. Text can have a foreground
+color, a background color and an intensity:
+
+.. code-block:: dylan
+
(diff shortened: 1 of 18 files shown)


sources/lib/coloring-stream/ansi-coloring-stream.dylan
```diff
@@ -0,0 +1,60 @@
...
+    (attributes :: <text-attributes>)
+ => (ansi-codes :: <byte-string>)
+  let fg = attributes.text-foreground-color;
+  let bg = attributes.text-background-color;
+  let intensity = attributes.text-intensity;
+  with-output-to-string(s)
+    write(s, "\<1b>[");
+    let need-separator = #f;
+    if (fg)
+      write(s, integer-to-string(30 + fg.color-ansi-code));
```
cgay: What are 30 and 40?  Worth using constants?

documentation/library-reference/source/coloring-stream/index.rst
```diff
@@ -0,0 +1,221 @@
...
+
+     Instances of ``<text-attributes>`` are used for representing the
+     desired text appearance. They can be passed to :gf:`format` when
+     writing to a :class:`<coloring-stream>`.
+
+     *background* and *foreground*, if given, should be one of the color
+     constants like :const:`$color-red`, :const:`$color-green`, etc.
+
+     *intensity*, if given, should be one of :const:`$bright-text-intensity`,
+     :const:`$dim-color-intensity` or :const:`$normal-text-intensity`.
```
cgay: should be ``$dim-text-intensity``
(two occurrences)

documentation/library-reference/source/coloring-stream/index.rst
```diff
@@ -0,0 +1,221 @@
...
+   :description:
+
+     .. note:: Not all terminals support dimmed text.
+
+.. constant:: $normal-text-intensity
+
+Text Colors
+-----------
+
+.. constant:: $color-black
```
cgay: What about leaving off the "color-" prefix? People can always rename on import in the unlikely case they're also defining ``$red`` etc.

waywardmonkeys: I've been thinking about this as well. One argument that is either for or against, I'm not sure which, is that DUIM and Deuce have constants named ``$red`` and so on.

documentation/library-reference/source/coloring-stream/index.rst
waywardmonkeys: I fixed this after your comment, but then I decided to rename the constants to ``$dim-intensity`` and so on.

waywardmonkeys:
I've updated this with some (more) bikeshedding.

I've also included a prototype commit that colorizes compiler warnings. (Clearly a work in progress.)

PR #951, opened 2015-08-03
[dfmc-conversion] Improve serious warnings which used reasons.

waywardmonkeys:
Some warnings had 'reasons' that explained in greater detail. These
resulted in extraneous double quotes being embedded in the warning
message.

We fix this by creating a warning subclass for each reason so that
they can be written individually and correctly.

In the future, we can extend these with further information as well
so that the causes for the warning can be laid out more clearly,
rather than saying something abstract like "something didn't match."

* ``sources/dfmc/conversion/define-domain-mop.dylan``
  (``<domain-types-not-valid>``): Make abstract.
  (``<domain-types-not-valid-required-argument-count>``,
   ``<domain-types-not-valid-required-argument-type>``): New subclasses.
  (``^domain-types-match-generic?``): Return a subclass which represents the
    correct warning rather than a reason string which was embedded in a
    generic warning.
  (``check-model`` on ``<&domain>``): Switch to using the returned subclass
    to create the warning.

* ``sources/dfmc/conversion/define-generic-mop.dylan``
  (``<method-not-congruent>``): Make abstract.
  (``<method-not-congruent-required-argument-count>``,
   ``<method-not-congruent-required-argument-type>``,
   ``<method-not-congruent-not-both-keyword>``,
   ``<method-not-congruent-not-both-variable>``,
   ``<method-not-congruent-mandatory-keyword>``,
   ``<method-not-congruent-required-values-count>``,
   ``<method-not-congruent-required-values-type>``,
   ``<method-not-congruent-generic-values-not-variable>``,
   ``<method-not-congruent-required-values-count-too-small>``,
   ``<method-not-congruent-no-required-values>``,
   ``<method-not-congruent-rest-values-type>``): Replace similarly named
    string constants with new subclasses.
  (``^signatures-congruent?``): Return a subclass which represents the
    correct warning rather than a reason string which was embedded
    in a generic warning.
  (``check-congruence``): Switch to using the returned subclass to create
    the warning.

### DIFF
--- a/sources/dfmc/conversion/define-domain-mop.dylan
+++ b/sources/dfmc/conversion/define-domain-mop.dylan
@@ -16,13 +16,23 @@ define program-warning <domain-visible-to-sibling-libraries>
     variable-name;
 end program-warning;
 
-define serious-program-warning <domain-types-not-valid>
+define abstract serious-program-warning <domain-types-not-valid>
   slot condition-variable-name,
     init-keyword: variable-name:;
-  slot condition-problem,
-    required-init-keyword: problem:;
-  format-string "Domain types for %= are not valid - %s.";
-  format-arguments variable-name, problem;
+  format-arguments variable-name;
+end serious-program-warning;
+
+define serious-program-warning <domain-types-not-valid-required-argument-type>
+    (<domain-types-not-valid>)
+  format-string "Domain types for %= are not valid - "
+      "some of the method's required parameter specializers "
+      "aren't subtypes of their counterparts in the generic.";
+end serious-program-warning;
+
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
The result of this is that a compiler warning like:

> /opt/opendylan-current/sources/io/streams/native-speed.dylan:76-77: Serious warning - Domain types for do-next-output-buffer are not valid - "some of the method's required parameter specializers aren't subtypes of their counterparts in the generic".

becomes:

> /opt/opendylan-current/sources/io/streams/native-speed.dylan:76-77: Serious warning - Domain types for do-next-output-buffer are not valid - some of the method's required parameter specializers aren't subtypes of their counterparts in the generic.

PR #956, opened 2015-08-15
[jamfiles] Clean up whitespace in mini-jambase.jam.

waywardmonkeys:
* sources/jamfiles/mini-jambase.jam: Clean up whitespace.


waywardmonkeys:
@housel: I did this as a PR in case you have feedback on cleaning this up. We already differ from upstream ``Jambase`` and doing a ``diff -w`` isn't going to make comparisons any more difficult.

PR #957, opened 2015-08-15
dfmc-modeling: Preserve <primitive> signature specifications

waywardmonkeys:
Preserve signature-spec information for <primitive> and <c-function>
models so that parameter names can be included in error messages
concerning them.

* sources/dfmc/modeling/raw-types.dylan
  (primitive-signature-spec): New slot added to <primitive>.

* sources/dfmc/modeling/modeling-library.dylan
  (module dfmc-modeling): Export primitive-signature-spec.

* sources/dfmc/conversion/define-primitive.dylan
  (compute-form-model-object on <primitive-definition>): Preserve the
   signature-spec when constructing the <primitive> model.

* sources/dfmc/conversion/convert-c-ffi.dylan
  (convert-%c-call-function, convert-%c-call-function-indirect,
   convert-%objc-msgsend): Preserve the signature-spec when
   constructing the respective model.

### DIFF
--- a/sources/dfmc/conversion/convert-c-ffi.dylan
+++ b/sources/dfmc/conversion/convert-c-ffi.dylan
@@ -91,6 +91,7 @@ define method convert-%c-call-function
            c-function-name: name & as-string(name),
            c-signature: ffi-signature,
            signature: signature,
+           signature-spec: sig-spec,
            c-modifiers: as-string(modifiers));
   convert-primitive-call(env, context, <primitive-call>, function, arguments);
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
@housel: This is one of your commits from your ``llvm-devel`` branch, but cleaned up a little to work on current master.

You used it to then provide a foundation for checking the primitive signatures at compile time to emit warnings. That part isn't ready to go in, but having this in place by default would be convenient.

PR #960, opened 2015-08-21
[configure] Handle MPS config for LLVM run-time.

waywardmonkeys:
While we don't yet support MPS for the LLVM run-time, we can at
least handle it correctly in an undocumented fashion.

* configure.ac: Validate MPS and set up the correct variables for
  both HARP and LLVM run-times when it has been requested.


waywardmonkeys:
Going to do this another way ...

PR #977, opened 2015-09-10
Start fixing issues with LLVM and MPS

waywardmonkeys:
This updates the build system as needed. It doesn't yet address compilation errors.


waywardmonkeys:
@housel: I'd like to get this on master so that it is easier if the Ravenbrook people should find the time to help out.

PR #979, opened 2015-09-12
[coloring-stream] Start to improve Windows support.

waywardmonkeys:
* sources/lib/coloring-stream/coloring-stream.dylan
  (coloring-stream-class-for-stream): On Windows, if running under
    ConEmu (like with cmder), return an ANSI stream, otherwise
    return a Windows coloring stream.
  (``<windows-coloring-stream>``): Define new class to anchor setting
    text attributes on Windows. This is here so that it is available
    everywhere to simplify code in coloring-stream-class-for-stream.

* sources/lib/coloring-stream/library.dylan
  (module coloring-stream-internals): Import $os-name.

* sources/lib/coloring-stream/windows-coloring-stream.dylan
  (print-object): Provide null implementation for now.

* sources/lib/coloring-stream/windows-coloring-stream.lid: Build
  windows-coloring-stream.dylan on Windows so that the implementation
  of print-object is available.

* sources/registry/x86-win32/coloring-stream: Point to the new
  windows-coloring-stream.lid.

### DIFF
--- a/sources/lib/coloring-stream/coloring-stream.dylan
+++ b/sources/lib/coloring-stream/coloring-stream.dylan
@@ -33,7 +33,15 @@ define method coloring-stream-class-for-stream
     (stream :: <file-stream>)
  => (coloring-stream-class :: <class>)
   if (stream-supports-color?(stream))
-    <ansi-coloring-stream>
+    if ($os-name == #"win32")
+      if (environment-variable("ConEmuANSI") = "ON")
+        <ansi-coloring-stream>
+      else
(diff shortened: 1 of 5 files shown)


waywardmonkeys:
@housel: This might help you. I haven't actually built / tested it yet.

waywardmonkeys:
@housel: I fixed both of your comments on the pull request. I also changed the comparison with ``$os-name`` to use ``==`` instead of ``=`` as it is a symbol.

PR #981, opened 2015-09-15
[dfmc-reader] Tighten types on fragment-source-location.

waywardmonkeys:
* sources/dfmc/reader/fragments.dylan
  (generic ``fragment-source-location``): This only accepts fragments now.
  (``fragment-source-location`` on ``<sequence>``): This is no longer used
    and has been removed.


waywardmonkeys:
This passed a 3 stage bootstrap. I inspected some of the call sites by hand to see if everything looked okay.

PR #985, opened 2015-09-28
alloca() inhabits <stdlib.h> in the BSD world

krytarowski:


waywardmonkeys:
Pushed this directly to the branch.

PR #987, opened 2015-10-02
documentation: Document library startup

housel:


documentation/hacker-guide/source/runtime/startup.rst
```diff
@@ -0,0 +1,138 @@
...
+- When :drm:`<symbol>` resolution is required, a call to
+  ``primitive-symbol-fixup`` is generated, passing a table of symbols
+  to resolve and reference addresses to patch.
+
+- If there are any system initializers in the compile unit, calls to
+  these are also generated. These normally only occur when a class
+  definition depends on non-static values.
+
+The remaining for-user initialization is done in the per-library
+startup glue function, named ``_Init_``\ *mangled-library-name*, where
```
waywardmonkeys: There are a variety of ``\`` scattered through the file ... are they for anything or a remnant from something else?

alf239: It looks like \TeX's unbreakable space — and, judging by the usage, it was supposed to be an unbreakable space.

alf239: Can you run `rst2latex` on the file? I haven't got RST tools at hand, so cannot check.

waywardmonkeys: Ah, yes, makes sense. Thanks!

PR #993, opened 2015-10-10
[dfmc-c-linker] Fix indirect C-functions.

### LINKED ISSUE #990: c-ffi - indirect functions - incorrect behavior when the c function requires no arguments.
```
from /opt/opendylan-2014.1/bin/../lib/libdfmc-c-linker.so
// If there's no C name, it's an indirect function and the
// first parameter is actually the function itself.
 //---*** Better: Add a slot to <&c-function> or a <&c-indirect-function>
copy-sequence(o.c-signature.^signature-required, start: 1)
```

This segment is only valid when at least 1 argument must be passed to the indirect C function. However, functions such as

```
//test.c
#include <stdio.h>
void display_stuff(void) {
 printf("hello, world!\n");
}
//end test.c
```
Have no parameters, causing a crash at compile-time.

### LINKED ISSUE #991: c-ffi - indirect functions - incorrect handling of function pointer.
Related to issue #990 the argument-list generated for the C function signature at compile-time does not match the list of arguments passed to the function pointer. One of the following actions are required to fix this bug, depending on what the function pointer actually points to (a call wrapper taking a pointer to the c function as a first argument, or the c function itself):
1- Require a `<c-function-pointer>` argument as the first parameter of a c-function definition. In that case, the argumnet-list (passed to the function pointer at the point of call) must also skip its first argumnent (cf. the copy-sequence from #990). 
2- Do not require a `<c-function-pointer>` argument as the first parameter of a c-function definition. In that case, the signature parameter-list shouldn't drop its first parameter, which is actually the first parameter of the `<c-function-pointer>` that will be passed at call-time, and NOT (as is assumed currently) the `<c-function-pointer>` itself.

waywardmonkeys:
Prior to this commit, indirect C-functions only worked on the C back-end
when they had a single parameter. If they had no parameters, then dylan-compiler
would error, and if they had 2 or more, then the generated C would be incorrect
and result in a C compiler error.

Fixes #990. Fixes #991.

* sources/dfmc/c-linker/c-link-c-object.dylan
  (emit-parameter-types): Emit all of the required parameters, even
   if it is indirect as the function pointer won't be present in this
   required parameter list. Also, handle having no required parameters
   and correctly generating "(void)" as the parameter list rather than
   "()" which means the wrong thing in C.

### DIFF
--- a/documentation/release-notes/source/2015.1.rst
+++ b/documentation/release-notes/source/2015.1.rst
@@ -112,6 +112,10 @@ Compiler
   used have been fixed. These are typically edge cases like ``+.5``
   or ``-3d3`` rather than commonly used literal notations.
 
+* The C back-end correctly handles indirect C-functions (where a
+  function pointer is given to be invoked rather than a direct
+  function call).
+
 Debugging
 =========
 
--- a/sources/dfmc/c-linker/c-link-c-object.dylan
+++ b/sources/dfmc/c-linker/c-link-c-object.dylan
@@ -8,20 +8,18 @@ Warranty:     Distributed WITHOUT WARRANTY OF ANY KIND
 define method emit-parameter-types
     (back-end :: <c-back-end>, stream :: <stream>, o :: <&c-function>) => ()
   format(stream, "(");
-  for (type in if (o.c-function-name)
-                 o.c-signature.^signature-required
-               else
-                 // If there's no C name, it's an indirect function and the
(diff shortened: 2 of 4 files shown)


waywardmonkeys:
Don't merge this yet. I plan to do a full bootstrap with it and also extend the ``c-ffi`` test suite a bit.

waywardmonkeys:
This should be good to go now. (@housel)

PR #1000, opened 2015-10-18
Open <mutable-object-with-elements>

waywardmonkeys:


waywardmonkeys:
I currently rely upon this in a library that I'm writing.

housel:
I've always been skeptical about `<object-with-elements>` but this is reasonable iff that is.

PR #1005, opened 2015-10-31
dylan: Fix size parameter handling in range

abeaumont:
Current implementation of range gives preference to size over to/below
parameters and modifies to accordingly if needed. This modification is
not correct as current as test range-5 in apple dylan test suite shows.

This change fixes size's behaviour according to current implementation,
making every test case in range-5 pass.

* sources/dylan/range.dylan
  (make): Fix size parameter handling

### DIFF
--- a/sources/dylan/range.dylan
+++ b/sources/dylan/range.dylan
@@ -75,14 +75,14 @@ define method make
     else
       if (size)
         if (to)
-          let new-to = from + by * (size + 1);
-          if (new-to <= to)
+          let new-to = from + by * (size - 1);
+          if (abs(new-to) <= abs(to))
             to := new-to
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
But the DRM says:

> If size: is specified and one of to:, above: or below: is specified, than it is an error if the number of elements implied by the to:, above: or below: argument (and the by: argument, if present) does not agree with the number of elements specified by the size: argument.

So is this fix really correct?

abeaumont:
I consider it a workaround, not a fix: it fits better with current implementation, and makes the testsuite pass. The idea is to make it raise an error as the DRM says, but we can have a sounder implementation in the meantime

PR #1006, opened 2015-10-31
Numbers redesign

abeaumont:
First 3 commits of the work towards a more extensible numerics are mergeable by themselves.

### DIFF
--- a/sources/dylan/integer.dylan
+++ b/sources/dylan/integer.dylan
@@ -241,23 +241,23 @@ define method \^
   end if
 end method \^;
 
-define function logior (#rest integers) => (logior :: <integer>)
+define sealed method logior (#rest integers) => (logior :: <integer>)
   reduce(binary-logior, 0, integers)
-end function logior;
+end method logior;
 
-define function logxor (#rest integers) => (logxor :: <integer>)
+define sealed method logxor (#rest integers) => (logxor :: <integer>)
   reduce(binary-logxor, 0, integers)
-end function logxor;
+end method logxor;
 
-define function logand (#rest integers) => (logand :: <integer>)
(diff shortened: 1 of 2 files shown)


sources/dylan/integer.dylan
```diff
@@ -241,23 +241,23 @@ define method \^
   end if
 end method \^;
 
-define function logior (#rest integers) => (logior :: <integer>)
+define sealed method logior (#rest integers) => (logior :: <integer>)
```
waywardmonkeys: Converting `logior` (and `logxor` and `logand`) to `sealed method` won't change anything, since they have no required arguments, so there is no dispatch happening.

sources/dylan/integer.dylan
```diff
@@ -272,16 +272,16 @@ define integer-binary-logical binary-logior machine-word-logior identity;
 define integer-binary-logical binary-logand machine-word-logand identity;
 define integer-binary-logical binary-logxor machine-word-logxor force-integer-tag;
 
-define inline function lognot (x :: <integer>) => (result :: <integer>)
+define sealed inline method lognot (x :: <integer>) => (result :: <integer>)
```
waywardmonkeys: Since this is a method now and not a function, should we explicitly specify a generic for it?

(This also goes for `logbit?`, `binary-logior`, `binary-logand`, `binary-logxor`, `ash`, `ash-right`, and `lsh`.)

sources/dylan/integer.dylan
xjswxdx: good

sources/dylan/integer.dylan
xjswxdx: ok

xjswxdx (commented):
you

PR #1009, opened 2015-11-07
Changes to transcendentals

waywardmonkeys:
This adds ``ilog2``, reduces the sealing on some functions, and removes a specialization that was overly general.


waywardmonkeys:
Updated per comment from @housel.

PR #1021, opened 2016-01-14
Threads cleanup

waywardmonkeys:
This is a precursor to moving the thread ID changes from the `current-thread-id` branch into the runtime code.

### DIFF
--- a/sources/dylan/thread.dylan
+++ b/sources/dylan/thread.dylan
@@ -30,6 +30,8 @@ define sealed class <thread> (<portable-double-container>)
 
 end class;
 
+ignore(priority);
+
 define sealed class <synchronous-thread> (<thread>)
 end class;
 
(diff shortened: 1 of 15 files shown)


sources/dylan/thread.dylan
```diff
@@ -30,6 +30,8 @@ define sealed class <thread> (<portable-double-container>)
 
 end class;
 
+ignore(priority);
```
vendethiel: I'm not sure I understand what this does outside of a class, could you explain? Thanks! (I'd expect it to ignore an argument, kinda CL-style)

waywardmonkeys: @vendethiel 

It does ignore the argument, CL-style. It works because a slot declaration creates a generic function (well, a mutable slot generates 2: ``priority`` and ``priority-setter``, a ``constant slot`` just generates the first).

Since nothing on the Dylan side is using that ``priority`` slot, we get a warning.

I thought about renaming the slot to ``thread-priority`` and then exporting it, but that seemed like a good step for another time if it is needed.

vendethiel: I see, thanks!

PR #1027, opened 2016-01-21
[llvm run-time] Don't use sys_errlist.

waywardmonkeys:
``sys_errlist`` has long been deprecated and isn't supported on all
platforms.

* sources/lib/run-time/llvm-posix-threads.c
  (``primitive_wait_for_simple_lock_timed``): Use ``strerror``
   instead of ``sys_errlist``.


waywardmonkeys:
Notably, any platform using musl libc, like emscripten, has no ``sys_errlist``.

PR #1033, opened 2016-01-27
[dylan] Remove overflow checks in linear-class-key-lookup

waywardmonkeys:
The loop in linear-class-key-lookup is known to be correct and
we don't need overflow checks when we're getting the new index
values. Removing this overflow turns the inner loop from:

    00000210  cmpl  %edi, %esi
    00000212  je  0x226
    00000214  movl  %esi, %ebx
    00000216  andl  $-0x4, %ebx
    00000219  cmpl  %ecx, 0x1c(%eax,%ebx)
    0000021d  je  0x22e
    0000021f  addl  $0x8, %esi
    00000222  jno 0x210

into:

    00000210  leal  0x1(%edi), %ebx
    00000213  andl  $-0x4, %ebx
    00000216  cmpl  %ecx, 0x1c(%edx,%ebx)
    0000021a  je  0x22b
    0000021c  addl  $0x8, %edi
    0000021f  cmpl  %edi, %esi
    00000221  jne 0x210

Here, we avoid the jno and perform the loop check at the end
of the loop rather than the start (with the LLVM back-end).

This helps lay the ground for some further optimizations in
subsequent commits.

* sources/dfmc/modeling/namespaces.dylan
  (&module machine-word-lowlevel): Export machine-word-add and
   machine-word-subtract so that other modules may use them.

* sources/dylan/new-dispatch.dylan
  (add-without-overflow): New helper function.
  (linear-class-key-lookup): Call add-without-overflow to avoid
   integer overflow checks.

### DIFF
--- a/sources/dfmc/modeling/namespaces.dylan
+++ b/sources/dfmc/modeling/namespaces.dylan
@@ -2273,6 +2273,8 @@ define &module machine-word-lowlevel
          machine-word-count-low-zeros,
          machine-word-count-high-zeros;
   // arithmetic
+  create machine-word-add,
+         machine-word-subtract;
   create machine-word-add-with-overflow,
          machine-word-subtract-with-overflow,
          machine-word-multiply-with-overflow,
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This doesn't address the bug where we don't stay in linear scans when we should and we prematurely punt to hash-based lookups. That'll be a separate pull request.

PR #1039, opened 2016-02-08
[llvm] Add missing intrinsics for copysign, clear_cache.

waywardmonkeys:
These intrinsics are in LLVM 3.5.

* sources/lib/llvm/llvm-intrinsics.dylan: Add intrinsic makers
  for copysign and clear_cache.

### DIFF
--- a/sources/lib/llvm/llvm-intrinsics.dylan
+++ b/sources/lib/llvm/llvm-intrinsics.dylan
@@ -162,6 +162,22 @@ begin
               linkage: #"external");
        end;
 
+  $llvm-intrinsic-makers["llvm.clear_cache"]
+    := begin
+         let function-type
+           = make(<llvm-function-type>,
+                  return-type: $llvm-void-type,
+                  parameter-types: vector($llvm-i8*-type, $llvm-i8*-type),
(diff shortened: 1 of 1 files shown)


housel:
I was going to suggest running llvm-tablegen on the LLVM 3.5 Intrinsics.td file, but I discovered that it uses a syntactic construct that my tablegen subset doesn't like. I'll see how involved it would be to fix.

waywardmonkeys:
That could be useful. I was going to submit a patch after this one that added everything since 3.5:

* llvm.assume
* llvm.masked.load.*
* llvm.masked.store.*
* llvm.minnum.*
* llvm.maxnum.*
* llvm.instrprof_increment
* llvm.canonicalize.*
* llvm.masked.gather.*
* llvm.masked.scatter.*
* llvm.bitset.test
* llvm.get.dynamic.area.offset.*
* llvm.instrprof_value_profile
* llvm.bitreverse.*
* llvm.invariant.group.barrier

waywardmonkeys:
@housel handled this via ``llvm-tablegen`` in c7e7d27e1bd01ad2eca02ec4844540332fc58d19

Issue #1043, opened 2016-02-14
type check error from mv from function without enough required values


waywardmonkeys:
This is a variant of #363, but I do not think the cause is the same. (I have a fix for #363, but it doesn't fix this.)

The code to reproduce this is:

```dylan
define function foo () => (a)
  1
end;

define function main (name :: <string>, arguments :: <vector>) => (a :: <string>)
  let (ign, ore, maybe-a :: false-or(<string>)) = foo();
  let a :: <string> = maybe-a | "a";
  a
end function main;

main(application-name(), application-arguments());
```

That results in this compile-time warning:

```
destructuring-bind.dylan:11-13: Warning - Type check can fail - singleton(#f :: <boolean>) inferred, <string> expected.
        ------------------------------------------------------
  11    let (ign, ore, maybe-a :: false-or(<string>)) = foo();
  12    let a :: <string> = maybe-a | "a";
  13    a
      ---
```

One interesting thing to note is that if you replace this line:

```dylan
  let a :: <string> = maybe-a | "a";
```

with this one:

```dylan
  let a :: <string> = #f | "a";
```

The warning goes away. In theory, these 2 bits of code should be identical since we should know that ``maybe-a`` is ``#f``.

The type estimate for ``maybe-a`` is coming from an ``<extract-single-value>`` from ``type-inference-rules type-infer-multiple-values`` in https://github.com/dylan-lang/opendylan/blob/c7e7d27e1bd01ad2eca02ec4844540332fc58d19/sources/dfmc/typist/typist-inference.dylan#L930-L969

waywardmonkeys:
@BarAgent You might find this interesting as this is one of the causes for the warnings that pop up from the ``peg-parser`` code when it is used.

In your case, this happens in the lexeme methods in the pl0 example due to ``read-matching-string`` having a single return value, but getting a destructuring bind used to extract 3 values.

waywardmonkeys:
It might be that ``dfmc-optimize`` should be constant folding some stuff away here that it is not ... such as when extracting a single value from a ``<multiple-value-temporary>`` which is generated by a ``<simple-call>``.

In the case that I was looking at in ``peg-parser`` which led to this bug report, I saw this under ``dylan-lldb``:

```
(lldb) expr -P1 -- (dylan_value)0x03dcb520
(dylan_value) $13 = 0x03dcb520 {<multiple-value-temporary>} {
  [users] = 0x048af580 {<pair>}
[...]

PR #1044, opened 2016-02-15
[dfmc-conversion] Fix destructuring bind.

### LINKED ISSUE #363: type inferencing error in destructuring bind
````dylan
define function main (name :: <string>, arguments :: <vector>) => (a :: <string>)
  let (ign, ore, maybe-a :: false-or(<string>)) = #t;
  let a :: <string> = maybe-a | "a";
  a
end function main;
````

This results in the warning:

````
Type check can fail - singleton(#f :: <boolean>) inferred, <string> expected.
````

This has been extracted from some code in the ``peg-parser`` by @BarAgent at the request of @hannesm

### PR DESCRIPTION
**Pull request by waywardmonkeys**
When the ``<values>`` doesn't have enough elements in it, it is supposed
to be padded with references to the ``#f`` object. Instead, it was writing
with the wrong loop index, so the ``<values>`` object was incorrect:

      *t3(3) := [VALUES ^#t ^#f #f]

Note that the last element is ``#f`` and not ``^#f``.

This resulted in confusing type check errors at compile time as detailed
in issue #363 where this code:

    let (ign, ore, maybe-a :: false-or(<string>)) = #t;
    let a :: <string> = maybe-a | "a";

Would result in a warning:

    Type check can fail - singleton(#f :: <boolean>) inferred, <string> expected.

This was found while minimizing a test case from the peg-parser, but is
not the same bug as the peg-parser is experiencing.

Fixes #363.

* sources/dfmc/conversion/convert.dylan
  (``pad-multiple-values``): Write ``#f`` object references at the correct
   index into the vector. (Use the right loop index value, ``j``, instead
   of ``i``.)

### DIFF
--- a/sources/dfmc/conversion/convert.dylan
+++ b/sources/dfmc/conversion/convert.dylan
@@ -386,7 +386,7 @@ define function pad-multiple-values
     mv[i] := ref;
   finally
     for (j :: <integer> from i below max-size)
-      mv[i] := make-object-reference(#f);
+      mv[j] := make-object-reference(#f);
     end for;
   end for;
   mv

### CODE REVIEW

**Comment by waywardmonkeys** [comment]
It is worth noting that I have no idea what all why this code generates that type check warning:

```
let (ign, ore, maybe-a :: false-or(<string>)) = #t;
let a :: <string> = maybe-a | "a";
```

Since we know that ``maybe-a`` is a ``false-or(<string>)`` (and actually, here we know that it is ``#f``), then it doesn't make sense that it would generate a warning for determining the value of ``a`` since it is equivalent to ``let a :: <string> = if (maybe-a) maybe-a else "a" end;`` ...

We don't see that happen here once we fix the bug that this pull request fixes because enough things get constant folded away ...  (I'm not worried about tracking that bug as part of this one ... it is related to what I'll solve when fixing issue #1043.

PR #1053, opened 2016-03-20
Range fixes

abeaumont:


sources/dylan/apple-dylan-test-suite/test-range.dylan
```diff
@@ -50,11 +50,11 @@ end test range-4;
 // not except through and up-to in a single expression
 //
 define test range-5 (description: "Through:, by:, and size:")
```
waywardmonkeys: This used to test ``through:`` but now tests ``to:``, right?

So it probably makes sense to:

1. Test these error conditions.
2. Make sure that the correct cases are tested as well and return the correct set of values.

waywardmonkeys:
The first commit has a typo in the commit message, "handlign", twice.

sources/dylan/apple-dylan-test-suite/test-range.dylan
abeaumont: That's what that description implies, but there's no such ``through:`` keyword anymore. I'll update the description and remove the comments above the test too, which don't apply anymore.

I'll add also correct cases too

abeaumont:
Fixed the commit message typo

sources/dylan/apple-dylan-test-suite/test-range.dylan
abeaumont: Done

waywardmonkeys:
Thanks!

Issue #747, opened 2014-05-16
Merge bsd-variable-search.lid and linux-variable-search.lid


waywardmonkeys:
The only difference between these files is the ``C-Libraries: -ldl`` in the Linux version. This isn't needed as the ``dylan`` library links against ``-ldl``.

Steps:

* Rename ``bsd-variable-search.lid`` to ``unix-variable-search.lid``
* Remove ``linux-variable-search.lid``
* Remove all of the Linux, FreeBSD and Darwin registry entries and just have ``registry/generic/variable-search`` point to the ``unix-variable-search.lid``
* Make sure everything still works on Linux by doing a full bootstrap.

waywardmonkeys:
Fixed by #1055.

PR #1055, opened 2016-03-24
[variable-search] Unify bsd and linux LID files

abeaumont:
The only difference between these files is the C-Libraries: -ldl in the
Linux version. This isn't needed as the dylan library links against
-ldl.

* sources/lib/variable-search/bsd-variable-search.lid:
  Rename to unix-variable-search.lid
* sources/lib/variable-search/linux-variable-search.lid: Remove
* sources/registry/generic/variable-search:
  Add registry file and point to point to the unix-variable-search.lid
* sources/registry/arm-linux/variable-search: Remove
* sources/registry/x86-freebsd/variable-search: Remove
* sources/registry/x86-linux/variable-search: Remove
* sources/registry/x86-netbsd/variable-search: Remove
* sources/registry/x86_64-freebsd/variable-search: Remove
* sources/registry/x86_64-linux/variable-search: Remove
* sources/registry/x86_64-netbsd/variable-search: Remove


abeaumont:
This should fix #747

PR #1058, opened 2016-04-18
[common-dylan] Add hexstring and from-hexstring to byte-vector.

wvdschel:
- sources/common-dylan/library.dylan
  (module byte-vector): Export hexstring and from-hexstring.
- sources/common-dylan/byte-vector.dylan
  Added hexstring and from-hexstring methods.

### DIFF
--- a/sources/common-dylan/byte-vector.dylan
+++ b/sources/common-dylan/byte-vector.dylan
@@ -222,7 +222,7 @@ define sealed inline method byte-storage-offset-address
               (the-buffer, primitive-repeated-slot-offset(the-buffer)))))
 end method;
 
-define method from-hexstring (string :: <byte-string>)
+define sealed method from-hexstring (string :: <byte-string>)
   => (result :: <byte-vector>)
   if (odd?(string.size))
     error("String size must be multiple of 2.");
@@ -235,7 +235,7 @@ define method from-hexstring (string :: <byte-string>)
       let hi :: <byte> = as(<byte>, string[i]);
       let lo :: <byte> = as(<byte>, string[i + 1]);
       let b  :: <byte> = ash(logand(hi, 15) + 9 * ash(hi, -6), 4) +
-        logand(lo, 15) + 9 * ash(lo, -6);
+                             logand(lo, 15) + 9 * ash(lo, -6);
       result[floor/(i, 2)] := b;
     end for;
   end without-bounds-checks;
@@ -243,7 +243,7 @@ define method from-hexstring (string :: <byte-string>)
   result;
 end method;
 
-define method hexstring (data :: <byte-vector>)
+define sealed method hexstring (data :: <byte-vector>)
   => (result :: <byte-string>)
   let result :: <byte-string> = make(<byte-string>, size: data.size * 2);
 
--- a/sources/common-dylan/byte-vector.dylan
+++ b/sources/common-dylan/byte-vector.dylan
@@ -221,3 +221,42 @@ define sealed inline method byte-storage-offset-address
            (primitive-repeated-slot-as-raw
               (the-buffer, primitive-repeated-slot-offset(the-buffer)))))
 end method;
+
+define method from-hexstring (string :: <byte-string>)
+  => (result :: <byte-vector>)
+  if (odd?(string.size))
+    error("String size must be multiple of 2.");
+  end if;
+
+  let result :: <byte-vector> = make(<byte-vector>, size: floor/(string.size, 2));
+
+  without-bounds-checks
+    for (i :: <integer> from 0 below string.size - 1 by 2)
+      let hi :: <byte> = as(<byte>, string[i]);
+      let lo :: <byte> = as(<byte>, string[i + 1]);
+      let b  :: <byte> = ash(logand(hi, 15) + 9 * ash(hi, -6), 4) +
+        logand(lo, 15) + 9 * ash(lo, -6);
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
I think it is worth soliciting ideas for the names of these functions (@cgay, @housel, @BarAgent?).

Also, the documentation should be updated, some simple tests added, and something added to the release notes.

sources/common-dylan/byte-vector.dylan
```diff
@@ -221,3 +221,42 @@ define sealed inline method byte-storage-offset-address
            (primitive-repeated-slot-as-raw
               (the-buffer, primitive-repeated-slot-offset(the-buffer)))))
 end method;
+
+define method from-hexstring (string :: <byte-string>)
```
cgay: I think we generally use x-to-y for conversions. In many (most?) cases it makes sense given that methods don't have a distinguished receiver. hex-string-to-byte-vector is quite a mouthful so perhaps just hex-to-bytes given that it's probably a reasonable assumption that "hex" implies string. and that "bytes" implies byte vector?

Is there any reason this method shouldn't accept ``<string>`` rather than ``<byte-string>``? Both byte strings and unicode strings (should they ever exist) may contain non-hex characters.

sources/common-dylan/byte-vector.dylan
```diff
@@ -221,3 +221,42 @@ define sealed inline method byte-storage-offset-address
...
+  end if;
+
+  let result :: <byte-vector> = make(<byte-vector>, size: floor/(string.size, 2));
+
+  without-bounds-checks
+    for (i :: <integer> from 0 below string.size - 1 by 2)
+      let hi :: <byte> = as(<byte>, string[i]);
+      let lo :: <byte> = as(<byte>, string[i + 1]);
+      let b  :: <byte> = ash(logand(hi, 15) + 9 * ash(hi, -6), 4) +
+        logand(lo, 15) + 9 * ash(lo, -6);
```
cgay: This line should be indented to the same level as the first "ash" on the previous line. I would state the general rule as: no part of an expression should be to the left of where the expression started.

sources/common-dylan/byte-vector.dylan
```diff
@@ -221,3 +221,42 @@ define sealed inline method byte-storage-offset-address
...
+    error("String size must be multiple of 2.");
+  end if;
+
+  let result :: <byte-vector> = make(<byte-vector>, size: floor/(string.size, 2));
+
+  without-bounds-checks
+    for (i :: <integer> from 0 below string.size - 1 by 2)
+      let hi :: <byte> = as(<byte>, string[i]);
+      let lo :: <byte> = as(<byte>, string[i + 1]);
+      let b  :: <byte> = ash(logand(hi, 15) + 9 * ash(hi, -6), 4) +
```
cgay: When working with hex I think it makes sense to use hex notation in the logand calls. e.g., ``logand(hi, #x0f)``

sources/common-dylan/byte-vector.dylan
```diff
@@ -221,3 +221,42 @@ define sealed inline method byte-storage-offset-address
...
+define method from-hexstring (string :: <byte-string>)
+  => (result :: <byte-vector>)
+  if (odd?(string.size))
+    error("String size must be multiple of 2.");
+  end if;
+
+  let result :: <byte-vector> = make(<byte-vector>, size: floor/(string.size, 2));
+
+  without-bounds-checks
+    for (i :: <integer> from 0 below string.size - 1 by 2)
```
cgay: I think it would be faster to maintain an index into result here as well, instead of calculating `floor/(i, 2)` each time around. Not sure though...I suppose the compiler might optimize it into a shift.

sources/common-dylan/byte-vector.dylan
wvdschel: The reason for using <byte-string> instead of <string> is efficiency - I originally used <string> but @waywardmonkeys instructed me to try changing it to <byte-string>.

The resulting DFM & generated C code is much more efficient, having fewer dispatches.

sources/common-dylan/byte-vector.dylan
wvdschel: My indenting is done using dylan-mode for emacs. I'm fine with correcting this, but is there a way I can instruct emacs to follow this general rule?

sources/common-dylan/byte-vector.dylan
wvdschel: All `floor/(..., 2)` calls in these functions are compiled to shifts:

    t130 := [PRIMOP machine-word-shift-right(t129, ^%2)]

I don't think keeping another index and incrementing it each time will be (significantly) faster.

sources/common-dylan/byte-vector.dylan
cgay: There are hacks, like using begin/end or () might work, but I don't know how to fix dylan-mode for real.

sources/common-dylan/byte-vector.dylan
cgay: Cool

wvdschel: https://www.youtube.com/watch?v=XwozVKOkzz4

wvdschel:
I've added some documentation and a release note.

When you decide on a name for these methods, let me know.

When there are no outstanding remarks, I will squash these commits into one.

PR #1086, opened 2018-05-20
Several small changes to run-time

### LINKED ISSUE #1066: Possible problem with LLVM bit code emission
Background: compiling gives some warnings about undefined behaviour (mostly left-shifting negative values, e.g. `I(-1)`) so I enabled clang's undefined behaviour sanitiser. To my surprise the build now failed when compiling the output of the llvm-runtime-generator.
````
clang -arch x86_64 -fexceptions -c -o obj-x86_64-darwin-llvm/x86_64-darwin-runtime.o x86_64-darwin-runtime.bc
error: Invalid alignment value
1 error generated.
````
The align parameter in the bitcode is calculated by https://github.com/dylan-lang/opendylan/blob/102d13bcbc80215e1fbf7b679bf36867c9733bf6/sources/lib/llvm/llvm-bitcode.dylan#L2398 which calls https://github.com/dylan-lang/opendylan/blob/0299e1f78b10b1acd8a9a70a1332b294e568f4f1/sources/common-dylan/numerics.dylan#L13 which calls https://github.com/dylan-lang/opendylan/blob/0299e1f78b10b1acd8a9a70a1332b294e568f4f1/sources/common-dylan/machine-words/logicals.dylan#L112 which calls https://github.com/dylan-lang/opendylan/blob/0299e1f78b10b1acd8a9a70a1332b294e568f4f1/sources/lib/run-time/run-time.h#L1613 
Now, `alignment-encoding` is sometimes called with 0 (I suppose the intention is to specify 'alignment determined by the ABI') and, I can see from the debugger that this gives crazy numbers for the alignments, because (I think) __builtin_clzl(0) is itself undefined. 
This could be fixed in alignment-encoding or any of the functions I've listed above, depends whether, in Dylan, any of these operations are 'undefined' in the C sense. On the other hand, the true problem could be calling `alignment-encoding(0)` since the parameter is a `false-or(<integer>)` and I believe `alignment-encoding(#f)` is the way to request default alignment.
What do you think?

### PR DESCRIPTION
**Pull request by pedro-w**
This PR gets rid of a few warnings when compiling the run-time library. It also fixes one test fail ("run-application input stream" in the system test suite) and one problem with LLVm code gen (fixes #1066)

### DIFF
--- a/sources/lib/run-time/c-run-time.c
+++ b/sources/lib/run-time/c-run-time.c
@@ -21,6 +21,10 @@
 #  define unlikely(x) x
 #endif
 
+/// Many of the short functions in this file are not used in the runtime
+//  but have been retained so that they can be called from within
+//  the debugger.
+
 /*
  stack allocation
    relies on inlining
@@ -531,15 +535,15 @@ dylan_value primitive_raw_as_string (DBSTR buffer) {
 
 /* SIGNATURES */
 
-INLINE dylan_simple_object_vector* signature_required(dylan_signature* sig) {
+dylan_simple_object_vector* signature_required(dylan_signature* sig) {
   return(sig->required);
 }
 
-INLINE dylan_simple_object_vector* signature_values(dylan_signature* sig) {
+dylan_simple_object_vector* signature_values(dylan_signature* sig) {
   return(sig->values);
 }
 
-INLINE dylan_value signature_rest_value(dylan_signature* sig) {
+dylan_value signature_rest_value(dylan_signature* sig) {
   return(sig->rest_value);
 }
 
@@ -553,41 +557,41 @@ INLINE dylan_value signature_rest_value(dylan_signature* sig) {
 #define REST_VALUE_P_MASK    0x080000
 #define NEXT_P_MASK          0x100000
 
-INLINE int signature_number_required(dylan_signature* sig) {
+int signature_number_required(dylan_signature* sig) {
   return(R(sig->properties) & NUMBER_REQUIRED_MASK);
 }
 
-INLINE int signature_number_values(dylan_signature* sig) {
+int signature_number_values(dylan_signature* sig) {
   return((R(sig->properties) & NUMBER_VALUES_MASK)
            >> NUMBER_VALUES_OFFSET);
 }
 
-INLINE int signature_key_p(dylan_signature* sig) {
+int signature_key_p(dylan_signature* sig) {
   return((R(sig->properties) & KEY_P_MASK) > 0);
 }
 
-INLINE int signature_all_keys_p(dylan_signature* sig) {
+int signature_all_keys_p(dylan_signature* sig) {
   return((R(sig->properties) & ALL_KEYS_P_MASK) > 0);
 }
 
-INLINE int signature_rest_p(dylan_signature* sig) {
+int signature_rest_p(dylan_signature* sig) {
   return((R(sig->properties) & REST_P_MASK) > 0);
 }
 
-INLINE int signature_optionals_p(dylan_signature* sig) {
+int signature_optionals_p(dylan_signature* sig) {
   return((R(sig->properties) & OPTIONALS_P_MASK) > 0);
 }
 
 
-INLINE int signature_rest_value_p(dylan_signature* sig) {
+int signature_rest_value_p(dylan_signature* sig) {
   return((R(sig->properties) & REST_VALUE_P_MASK) > 0);
 }
 
-INLINE int signature_next_p(dylan_signature* sig) {
+int signature_next_p(dylan_signature* sig) {
   return((R(sig->properties) & NEXT_P_MASK) > 0);
 }
 
-INLINE dylan_value signature_make_properties
+dylan_value signature_make_properties
     (int number_required, int number_values,
      int key_p, int all_keys_p, int rest_p, int rest_value_p) {
   return(I(number_required
@@ -600,7 +604,7 @@ INLINE dylan_value signature_make_properties
 
 /* FUNCTION */
 
-INLINE DFN function_xep(dylan_simple_method* function) {
+DFN function_xep(dylan_simple_method* function) {
   return(function->xep);
 }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by cgay** [comment]
I don't know much about the run-time so it would be best for @waywardmonkeys  or @housel  to look it over.

We generally tried not to have merge commits, avoidable if you --rebase when pulling to your master branch. (I'm just letting your know how things have been done; I'm not worried about a few merge commits in this PR.)

**Comment by pedro-w** [comment]
If those gentlemen could comment that would be great. Then I will re-do this PR without merges (not sure what happens in github terms if make a branch, and master is updated susequently.  I thought it better to keep it up-to-date)
The background to change d16cd660720f124c56b2426a355fbaf2605cad31 is http://colabti.org/irclogger/irclogger_log/dylan?date=2018-04-18#l7

**Comment by waywardmonkeys** [comment]
I will comment in the next 16 hours, I hope.

File: sources/lib/run-time/llvm-runtime.h [hunk 1]
```diff
@@ -18,7 +18,7 @@ typedef float                   DSFLT;
 typedef double                  DDFLT;
 
 #define ITAG 1
-#define I(n) ((dylan_value) (((intptr_t)(n) << 2) | ITAG))
+#define I(n) ((dylan_value) (((unsigned long)(n) << 2) | ITAG))
```
pedro-w: Actually I do think it might be good have an 'audit' of what C types we use here and elsewhere to check it's all sound with respect to 32 vs 64 bit. And maybe use the `<stdint.h>` types more?

pedro-w:
Would be great if someone could find a few moments to look over this, thanks in advance.

sources/lib/run-time/c-run-time.c
```diff
@@ -632,35 +583,11 @@ INLINE int function_number_required(dylan_simple_method* function) {
...
 
 
-INLINE int function_rest_value_p(dylan_simple_method* function) {
-  return(signature_rest_value_p(function->signature));
-}
-
-INLINE int function_next_p(dylan_simple_method* function) {
-  return(signature_next_p(function->signature));
-}
-
```
waywardmonkeys: These and many of the other unused functions were useful under the debugger or for docs in how to access that data.

sources/lib/run-time/c-run-time.c
```diff
@@ -525,7 +525,7 @@ dylan_value primitive_raw_as_string (DBSTR buffer) {
   size_t len = strlen(buffer);
   size_t size = round_up_to_word(base_size + len + 1);
   dylan_value string = primitive_alloc_leaf_r(size, &KLbyte_stringGVKdW, len, 1);
-  memcpy(((dylan_byte_string*)string)->data, buffer, len);
+  memcpy(((dylan_byte_string*)string)->data, buffer, len + 1);
```
waywardmonkeys: What does the LLVM run-time do here?

sources/lib/run-time/posix-threads.c
```diff
@@ -512,8 +506,6 @@ dylan_value primitive_make_thread(dylan_value t, dylan_value f, DBOOL s)
   thread->handle1 = 0;       // runtime thread flags
   thread->handle2 = rthread; // runtime thread object
 
-  // param.sched_priority = priority_map(priority);
-
```
waywardmonkeys: This should get implemented not deleted.

sources/lib/run-time/run-time.h
```diff
@@ -1609,8 +1609,12 @@ extern DMINT primitive_unwrap_abstract_integer(dylan_value);
...
 #ifdef OPEN_DYLAN_COMPILER_GCC_LIKE
-#define primitive_machine_word_count_low_zeros(x) __builtin_ctzl(x)
-#define primitive_machine_word_count_high_zeros(x) __builtin_clzl(x)
+static inline DMINT primitive_machine_word_count_low_zeros(DMINT x) {
+  return x ? __builtin_ctzl(x) : (8 * sizeof(x));
+}
+static inline DMINT primitive_machine_word_count_high_zeros(DMINT x) {
+  return x ? __builtin_clzl(x) : (8 * sizeof(x));
+}
 #else
```
waywardmonkeys: This needs a test.

sources/lib/run-time/c-run-time.c
housel: It copies the string, including the NUL.

sources/lib/run-time/llvm-runtime.h
housel: Agree that moving towards more `<stdint.h>` types would be a good thing.

sources/lib/run-time/c-run-time.c
pedro-w: Good point. I can add a comment to that effect and remove the INLINE which will stop them being flagged as unused. Does that sound OK?

sources/lib/run-time/c-run-time.c
pedro-w: I'm pretty confident this is correct. This bug would only become apparent if you took in a string via the C interface and then used it again in a C function. I saw it in one of the tests where it gets the current shell via `getenv` and uses it to `exec` something.

sources/lib/run-time/run-time.h
pedro-w: Agreed, will do.

pedro-w: Actually I'd forgotten, there is one. It's "machine-words-protocol-function-test" in libraries-test-suite-app
https://github.com/dylan-lang/opendylan/blob/30b4a9192eecf4f5458c88f3901866af9dc1cf90/sources/common-dylan/tests/machine-words.dylan#L68-L98

sources/lib/run-time/posix-threads.c
pedro-w: Not sure how to proceed on this one. As far as I can see POSIX doesn't allow you to set the 'nice' of a thread (though Linux does) - in which case we'd have to go for the realtime schedulers (SCHED_FIFO or SCHED_RR). Would these require a privileged process to run? And, Macos doesn't have `pthread_setschedparam` at all, or at least some versions don't.
So maybe a Linux only implementation is the best way?

cgay: No reason you need to implement this as part of this pull request. Just undelete those lines for now?

pedro-w: OK I've reverted the change to that file now.

cgay:
This looks ready to merge, to me. Am I missing anything?

PR #1087, opened 2018-05-22
Test fixes for the threads library

pedro-w:
A couple of tests had no checks and so were marked as not implemented. Corrected some mistakes in the interface to pthreads which caused test fails/lockups in the test app.


sources/lib/run-time/posix-threads.c
```diff
@@ -938,6 +938,13 @@ dylan_value primitive_wait_for_semaphore_timed(dylan_value l, dylan_value m)
   while (semaphore->count <= 0) {
     int res = pthread_cond_timedwait(&semaphore->cond, &semaphore->mutex, &end);
     if (res == ETIMEDOUT) {
+      /* The mutex is aquired even if the wait times out.
```
cgay: Typo: acquired

pedro-w: Dunce's cap for me. Now fixed.

PR #1089, opened 2018-05-23
Fix typo in macro name

### LINKED ISSUE #1088: Typo in check macro
https://github.com/dylan-lang/opendylan/blob/ae592564a7d14797ae3012471015e65a754a8c49/sources/common-dylan/tests/threads/threads.dylan#L71
Sorry this should have been `check-no-errors` or alternatively `check-no-condition` - I missed the compiler warning in the middle of the output (which is quite copious)

pedro-w:
Fixes #1088

### DIFF
--- a/sources/common-dylan/tests/threads/threads.dylan
+++ b/sources/common-dylan/tests/threads/threads.dylan
@@ -68,7 +68,7 @@ end test;
 // Just run thread-yield.
 //
 define test yield-test(description: "thread-yield")
-  check-no-error("thread-yield works",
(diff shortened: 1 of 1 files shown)


cgay:
This is exactly why I want to reduce compiler output.  (It's great that you're looking at fixing some of the long-standing warnings.)

pedro-w:
I believe that if I can address (most of) the warnings and test fails (either by fixing the library or fixing the test) that will give a sound footing to go forward.

PR #1093, opened 2018-06-07
Work on remove for ranges

### LINKED ISSUE #1091: remove on <range> can return an invalid object
If you remove the last element from a range, the result is a range of size 0 but for which `empty?` returns `#f`. For example:
``` dylan
  let r = make(<range>, from: 1, to: 3);
  while(r)
    format-out("%s, size %d, empty? %s\n", r, size(r), empty?(r));
    if (size(r) > 0)
      r := remove(r, first(r))
    else
      r := #f;
    end
  end;
```
output:
```
{range 1 through 3 by 1}, size 3, empty? #f
{range 2 through 3 by 1}, size 2, empty? #f
{range 3 through 3 by 1}, size 1, empty? #f
{empty range}, size 0, empty? #f
```
(note the last line)

pedro-w:
This is a small change to make sure that removing the element from a one-element range results in a valid empty range (instance of `<empty-range>` in this implementation)
A test is also provided.
Fixes #1091

### DIFF
--- a/sources/dylan/tests/regressions.dylan
+++ b/sources/dylan/tests/regressions.dylan
@@ -240,6 +240,15 @@ define test issue-440 ()
               map(identity, cis), cis);
 end;
 
+define test issue-1091 ()
+  let test-range = range(from: 1, size: 1);
+  check-false("one element range is not empty", empty?(test-range)); 
+  check-equal("one element range has size 1", 1, size(test-range));
+  test-range := remove(test-range, 1);
+  check-true("empty range is empty", empty?(test-range)); 
+  check-equal("empty range has size 0", 0, size(test-range));
+end;
+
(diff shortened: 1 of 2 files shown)


sources/dylan/tests/regressions.dylan
```diff
@@ -240,6 +240,15 @@ define test issue-440 ()
               map(identity, cis), cis);
 end;
 
+define test issue-1091 ()
```
cgay: I realize this hasn't been the tradition in this file to date, but could you add this?
``// Issue 1091: remove on <range> can return an invalid object``
and/or a URL for the issue?

pedro-w: Done.

PR #1121, opened 2018-09-11
LLVM TableGen 7.x

housel:
These changes implement additional LLVM TableGen features for generating intrinsics information needed for the `llvm` library.

### DIFF
--- a/sources/app/llvm-tablegen/tablegen-parser.dylgram
+++ b/sources/app/llvm-tablegen/tablegen-parser.dylgram
@@ -1,6 +1,6 @@
 Module:       llvm-tablegen
 Author:       Peter S. Housel
-Copyright:    Original Code is Copyright 2009-2011 Gwydion Dylan Maintainers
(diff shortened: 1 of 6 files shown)


sources/app/llvm-tablegen/tablegen-parser.dylgram
```diff
@@ -14,19 +14,61 @@ end class;
 
 define variable *tablegen-overrides* = make(<string-table>);
 
+define function save-overrides() => (previous-overrides :: <string-table>)
+  let previous-overrides = *tablegen-overrides*;
+  // FIXME
```
pedro-w: By the way, what needs fixing here, does `shallow-copy` not work on tables?

housel: It turns out that shallow-copy does work on tables, so I changed the code to use it. Perhaps it did not work in 2010.

PR #1128, opened 2018-09-20
Re-generated magic numbers from macOS 32-bit SDK

pedro-w:
This was based on the SDK from Xcode 9.4.1. 
Some of the structures have changed from when this file was previously generated in 2014.
Note that the 64-bit version has not changed since it was last generated.


waywardmonkeys:
IIRC, there's a way to make sure you get the versioned struct or something so that the sizes don't change in the future...

PR #1139, opened 2018-10-19
Update to LLVM 7.0 debug metadata and bitcode

housel:
These changes update LLVM debug metadata and many aspects of LLVM assembly-language syntax and the bitcode format to be compatible with LLVM 7.0; previously LLVM 3.5 (approximately) was targeted.


pedro-w:
This is great. But also rather daunting, there's so much of it! Have you any ideas on how to go about testing/verifying the output is correct?

Issue #94, opened 2011-12-14
Building Applications with DUIM: Fix links


waywardmonkeys:
There are links within this document that are still broken (search for ".htm").  These need special anchors inserted and linked up.

waywardmonkeys:
The link anchors still need to be fixed.

fraya:
Could you provide me an example?

PR #1148, opened 2018-11-02
Building Applications with DUIM: Fix links #94.

fraya:
Fix broken links finished in '.htm' to '.html'.


cgay:
Thanks!

waywardmonkeys:
The anchors in these links need to be fixed as well, which is why I didn’t fix this the other times that I worked in all of these docs.

PR #1170, opened 2018-12-16
Implement run-time backtraces and other improvements

housel:
These changes implement printing a backtrace when a Dylan program is terminated with an error condition, and make other improvements that improve platform support.

### DIFF
--- a/sources/lib/run-time/demangle.c
+++ b/sources/lib/run-time/demangle.c
@@ -1,0 +1,136 @@
+#include "demangle.h"
+
+#include <limits.h>
+#include <stdbool.h>
+
+#define CONSTANT_PREFIX         'K'
+#define DYLAN_MODULE_SEP        'K'
+#define MODULE_SEP              'Y'
+#define LIBRARY_SEP             'V'
+#define ESCAPE_SEP              'Z'
+#define METHOD_MARK             'M'
+#define IEP_MARK                'I'
+
+#define ARRAY_LEN(x)            (sizeof(x)/sizeof((x)[0]))
+
+// This is adequate for IEPs, but could be improved for other things
+// See documentation/hacker-guide/source/runtime/mangling.rst
+int dylan_demangle(char *dest, size_t destsize, char *src)
+{
+  if (destsize == 0) {
+    return -1;
+  }
+  char *limit = dest + destsize - 1;
+
+  static const char * const dylan_abbrev_map[] = {
+    ['d'] = "dylan",
+    ['i'] = "internal",
+    ['p'] = "dylan-primitives",
+    ['e'] = "dylan-extensions",
+    ['c'] = "dylan-c-ffi",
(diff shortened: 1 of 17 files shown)


sources/lib/run-time/demangle.c
```diff
@@ -0,0 +1,137 @@
...
+#define MODULE_SEP              'Y'
+#define LIBRARY_SEP             'V'
+#define ESCAPE_SEP              'Z'
+#define METHOD_MARK             'M'
+#define IEP_MARK                'I'
+
+#define ARRAY_LEN(x)            (sizeof(x)/sizeof((x)[0]))
+
+// This is adequte for IEPs, but could be improved for other things
+int dylan_demangle(char *dest, size_t destsize, char *src)
```
cgay: How about linking to https://opendylan.org/documentation/hacker-guide/runtime/mangling.html here?

sources/lib/run-time/demangle.c
```diff
@@ -0,0 +1,137 @@
...
+            *dest++ = *library++;
+          }
+          module_seen = true;
+        }
+        if (c == METHOD_MARK) {
+          *dest++ = '#';
+        }
+        break;
+
+        break;
```
cgay: Extraneous break statement?

sources/lib/run-time/stack-walker.c
```diff
@@ -1,71 +1,299 @@
...
+
+      fprintf(stderr, "  %#jx\n", (uintmax_t) ip);
+    }
+
+    // On to the next enclosing frame
+    rc = unw_step(&cursor);
+  } while (rc > 0);
 }
 
 #else
```
cgay: Add a comment since so far away from the #ifdef?

pedro-w:
I had a look at this by starting from the master branch of dylan-lang/opendylan and pulling from your branch run-time-backtrace-1. I was using debian 9 64-bit with clang-7 installed.
`make check` was OK (no fails that don't happen for master as far as I can see)
When I installed it and built my own app, created with `make-dylan-app cr`, built with `dylan-compiler -back-end llvm -build cr` linking failed at the last step, citing missing functions like `sin`, `cos` from `libcommon-dylan.so`.
I added `-lm` to x86_64-linux-build.jam and it was OK.
I tried to trigger a back-trace but never saw anything, is this right?
``` dylan
Module: cr
Synopsis: 
Author: 
Copyright: 

define function it()
  error ("here is an error")
end function;
  
define function main
  (name :: <string>, arguments :: <vector>)
  format-out("We're OK,");
  it();
  format-out("Hello, world!\n");
  exit-application(0);
end function main;

main(application-name(), application-arguments());
```
output

    $ ./cr
    here is an error

pedro-w:
OK, as I mentioned on slack - my fault. Need to re-run autogen.sh before configure and make.

PR #1195, opened 2019-02-24
dfmc-reader: Parse DEP 0010 element otherwise syntax

housel:
This commit changes the Dylan infix parser to handle otherwise-based
default expressions for bracketed element references.

* sources/dfmc/reader/parser.dylgram
  (operand productions): Handle operand [ arguments ] otherwise leaf
   expression syntax, converting to a element(arguments, default: leaf)
   function call.

* sources/dfmc/reader/infix-parser.dylan: Regenerate from parser.dylgram

* sources/dylan/tests/collections.dylan (test-element): Add a test
  case for element otherwise syntax.


housel:
This construct is likely too error-prone, so it's best that it be tabled for now.

PR #1199, opened 2019-03-01
build: Clean tool-generated files on make clean

housel:
* Makefile.in (clean): Clean sources generated using parser-compiler
  or scepter when doing "make clean".


waywardmonkeys:
Are these always in the srcdir and not the builddir?

housel:
Yes, the compiler only knows to look for source files in the source directory, even generated sources, and so the tools generate their sources within $(srcdir).

PR #1201, opened 2019-03-03
Add more thorough tests for bit-vectors

pedro-w:
This PR is long-winded but straightforward. There were some unimplemented tests for bit-vectors (i.e. tests without any checks in them) and some that seemed to be missing altogether. The plan seemed to be run the basic tests (and, or, xor, etc.) on vectors which were 'tiny', 'small', 'huge' and 'multiple-word' sized. Tiny and small were the same so I made tiny to be 4 bits then we are testing vectors that are smaller than one byte. Huge vectors (117 bits) are larger than any current machine word and small vectors (14 bits) are intermediate. Multiple-word sized vectors remain 96 bits.

### DIFF
--- a/sources/collections/tests/bit-count.dylan
+++ b/sources/collections/tests/bit-count.dylan
@@ -19,7 +19,7 @@ end test;
 
 define test bit-count-tiny-vector ()
   let vector = make(<bit-vector>, size: $tiny-size);
-  let bits = list(1, 4, 6, 7, 9, 11, 13);
+  let bits = list(1);
   set-bits(vector, bits);
   check-equal("Count bits in tiny vector", bit-count(vector), size(bits));
   check-equal("Count zero bits in tiny vector",
@@ -28,6 +28,16 @@ define test bit-count-tiny-vector ()
     bit-count(vector, bit-value: 1), size(bits));
 end test;
 
+define test bit-count-small-vector ()
+  let vector = make(<bit-vector>, size: $small-size);
+  let bits = list(1, 4, 6, 7, 9, 11, 13);
+  set-bits(vector, bits);
+  check-equal("Count bits in small vector", bit-count(vector), size(bits));
+  check-equal("Count zero bits in small vector",
+    bit-count(vector, bit-value: 0), $small-size - size(bits));
+  check-equal("Count one bits in small vector",
+    bit-count(vector, bit-value: 1), size(bits));
+end test;
 
 define test bit-count-huge-vector ()
   let vector = make(<bit-vector>, size: $huge-size);
@@ -55,6 +65,7 @@ end test;
 define suite bit-count-suite (description: "Test bit-count")
   test bit-count-empty-vector;
   test bit-count-tiny-vector;
+  test bit-count-small-vector;
   test bit-count-huge-vector;
   test bit-count-multiple-word-sized-vector;
 end suite;
--- a/sources/collections/tests/bit-count.dylan
+++ b/sources/collections/tests/bit-count.dylan
@@ -62,7 +62,7 @@ define test bit-count-multiple-word-sized-vector ()
     bit-count(vector, bit-value: 1), size(bits));
 end test;
 
-define suite bit-count-suite (description: "Test bit-count")
+define suite bit-count-suite ()
   test bit-count-empty-vector;
   test bit-count-tiny-vector;
   test bit-count-small-vector;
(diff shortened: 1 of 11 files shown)


sources/collections/tests/bit-vector-xor.dylan
```diff
@@ -40,72 +40,293 @@ define test bit-vector-xor-tiny-vector
...
     bits5-xor-bits6, bits5-xor-bits6, bits5-xor-bits6, bits5-xor-bits6);
   test-xor-for-all-pad-values("Xor two vectors again, arguments reversed",
-    vector6, vector5,
+    $huge-size, vector6, vector5,
     bits6-xor-bits5, bits6-xor-bits5, bits6-xor-bits5, bits6-xor-bits5);
 
 end test;
 
+define test bit-vector-xor-multiple-word-vector
+    (description: "Test bit-vector-andc2 with multiple-word sized bit-vectors")
```
cgay: Testworks no longer supports the "description" attribute, so that can be removed here. Just use a comment. The description slot was never used for anything.

(OD doesn't warn for extra keywords passed to `make`, which is what this ends up doing.)

I suggest moving all the descriptions to comments in a separate commit. Doesn't have to be in this PR either.

sources/collections/tests/bit-vector-xor.dylan
```diff
@@ -40,72 +40,293 @@ define test bit-vector-xor-tiny-vector
...
+    bits5, bits5, bits5, bits5);
+
+  // Xor an all-one vector with a mixed vector and different pad values
+  //
+  test-xor-for-all-pad-values("Xor an all-one and mixed vector",
+    $multiple-word-size, vector3, vector5,
+      not-bits5, not-bits5, not-bits5, not-bits5);
+  test-xor-for-all-pad-values("Xor a mixed and all-one vector",
+    $multiple-word-size, vector5, vector3,
+      not-bits5, not-bits5, not-bits5, not-bits5);
```
cgay: There are a few cases of inconsistent indentation like this.

(I suspect if I hit <tab> on most of these lines it will indent to underneath the first double quote.)

sources/collections/tests/bit-vector-xor.dylan
pedro-w: Should I be replacing `check-*` with `assert-*` too?

sources/collections/tests/bit-vector-xor.dylan
pedro-w: emacs dylan-mode is a pain for re-indenting things when my attention is elsewhere.

sources/collections/tests/bit-count.dylan
```diff
@@ -62,7 +62,7 @@ define test bit-count-multiple-word-sized-vector ()
     bit-count(vector, bit-value: 1), size(bits));
 end test;
 
-define suite bit-count-suite (description: "Test bit-count")
```
cgay: This doesn't compile, does it?  I believe you at least need the `()` to be there.

cgay: Pretend I put that comment after the following line. Still not used to GitHub review UI.

pedro-w: Fixed, thanks.

sources/collections/tests/bit-vector-xor.dylan
cgay: Not necessarily.  The plan is to change the semantics of `assert-*` to abort the rest of the test if they fail, whereas `check-*` will retain the current behavior.  If you want the assertion to abort the test you could switch, but I'd leave it for a different PR.

Given this change of plans (i.e, when I introduced `assert-*` it was purely to have a replacement for `check-*` that could auto-generate the failure message based on the code) I'm not sure what to do about the fact that `assert-*` and `check-*` have different parameter lists.

sources/collections/tests/bit-vector-xor.dylan
```diff
@@ -119,9 +118,8 @@ define test bit-vector-xor-small-vector ()
   let vector6 = make(<bit-vector>, size: $small-size, fill: 1);
   unset-bits(vector6, not-bits6);
 
-  let bits5-xor-bits6
-    = intersection(union(bits5, bits6),
-                   compute-not-bits(intersection(bits5, bits6), $small-size));
+  let bits5-xor-bits6 = intersection(union(bits5, bits6),
+    compute-not-bits(intersection(bits5, bits6), $small-size));
```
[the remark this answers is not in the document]
pedro-w: I changed all those indents to what dylan-mode does. More than 600 lines were changed - quite a lot of the code previously had the 'unusual' style. Should I commit it? I think probably we should bite the bullet and make it more consistent, but it is quite a big change in formatting relative to its change in actual content.

PR #1203, opened 2019-03-04
dfmc-reader: Allow underscores in numeric literals

cgay:
This implements DEP-0011,
https://opendylan.org/proposals/dep-0011-numeric-literal-syntax.html

Aside from the lexer changes to implement the new syntax, there is one
change to parse-integer:dfmc-reader to catch <arithmetic-error> instead
of <error>, since this masked other errors encountered during development.


sources/dfmc/reader/lexer.dylan
```diff
@@ -1032,7 +1035,7 @@ define method parse-integer
     else
       genericnegative(repeat(start, 0));
     end if;
-  exception (overflow :: <error>)
+  exception (overflow :: <arithmetic-error>)
```
waywardmonkeys: Should this be part of a separate commit? (And the next couple of changes as well...)

sources/dfmc/reader/tests/literal-test-suite.dylan
```diff
@@ -17,9 +17,11 @@ define function verify-presentation
...
 
 define test binary-integer-literal-test ()
-  let f = read-fragment("#b010");
-  verify-literal(f, 2, <integer-fragment>);
-
+  verify-literal(read-fragment("#b10"), 2, <integer-fragment>);
+  verify-literal(read-fragment("#b010"), 2, <integer-fragment>);
+  verify-literal(read-fragment("#b1111_0000"), 240, <integer-fragment>);
+  assert-false(read-fragment("#b10_"));
+  assert-signals(<invalid-token>, read-fragment("#b_10"));
```
waywardmonkeys: Should also test with multiple underscores.

waywardmonkeys:
In addition to the other comments ... you might consider changing the parsing tests first and then just adding the tests for the underscore changes.

Also, you might look at the code that presents / formats fragments back into code and see how they handle all of this ... but I bet that that'll show off some new problems.

sources/dfmc/reader/lexer.dylan
cgay: Yes, good point. I'll clean it up in several places.

cgay: I'm just dropping this change, actually, since I can't remember what it was for, exactly, and I can't guarantee `<arithmetic-error>` was the only kind of error it was intended to catch.

sources/dfmc/reader/tests/literal-test-suite.dylan
cgay: Done.

cgay:
> Also, you might look at the code that presents / formats fragments back into code and see how they handle all of this ... but I bet that that'll show off some new problems.

It looks to me like the fragment presentation code is only used for presenting warnings (via `print-object`).  I like that the warning here displays the integer in the standard form while the code snippet displays the original source form.  I think this is ok?

```
abc.dylan:12: Serious warning - Invalid type for argument key1 in call to method keyword-function (key1 :: <string>) => (#rest results :: <object>):  singleton(1000000 :: <integer>) supplied, <string> expected.
        ---------------------------
        keyword-function(1_000_000);
        ---------------------------
```

PR #1205, opened 2019-03-09
Reorganize configuration and build files

housel:
These changes reorganize the auxiliary files used for configuration and build, and make various improvements to the configuration step, including fixing the detection of whether or not LLVM code generation is supported.


BarAgent:
How is OVERVIEW.txt outdated? The majority of it is describing contents of directories and libraries, and a bit about internal conventions.

housel:
It's outdated in that

- I changed the organization of directories and didn't update it
- We now have the Hacker's Guide which would be a better place for this sort of information

I can drop that particular commit for now if people want.

cgay:
> I can drop that particular commit for now if people want.

Sure. I can take a stab at integrating its info into the Hacker Guide, or per-directory README files. (When you asked about this in chat I was thinking of the CONTRIBUTING file, which probably could be deleted now.)

Issue #1105, opened 2018-06-30
Error building dylan-test-suite (constants-test)


pedro-w:
When building the libraries-test-suite-app I get a C compiler error
```
constants.c:550:22: 
error: integer literal is too large to be represented in any integer type

  T0 = (dylan_value) 0xFFFFFFFFFFFFFFFF8000000000000001L;
```
from compiling this  test code:
https://github.com/dylan-lang/opendylan/blob/f8e2579fe0a614c6dfebe1984ac94d2464b4bfba/sources/dylan/tests/constants.dylan#L22-L27

There are three parts to this
1. There's a problem with DFMC's compile-time arithmetic. 
2. This test shouldn't happen at compile-time anyway, it's supposed to test the runtime addition
3. If the test is made to happen at runtime (easy enough to do) it doesn't raise a condition

I can look into this, but just wanted some opinions first - is this Mac specific or does everyone get this C error? And, on a bigger point, should numerical overflow be raising conditions? At the moment AFAICS it doesn't but implementing it would introduce overhead to many mathematical operations.
Possibly related: #78

cgay:
Did you mean https://github.com/dylan-lang/opendylan/issues/76 ?  Also these two seems in the same general area:
https://github.com/dylan-lang/opendylan/issues/863
https://github.com/dylan-lang/opendylan/issues/633

PR #1211, opened 2019-03-19
Make test-suite compile for C back end

pedro-w:
At the moment, compiling the `libraries-test-suite-app` fails because the C back end creates a constant that is too big for the compiler to handle (reported as #1105)
This PR is not a fix for that but it does make the arithmetic happen at run-time, which means building the test-suite completes successfully, although the test itself does still fail 

    dylan-extensions-protocol-constants-test failed
      $maximum-integer + 1 overflows failed [no condition signaled]
      $minimum-integer - 1 overflows failed [no condition signaled]

The proper fix for #1105 will take a bit more thought, I believe.

### DIFF
--- a/sources/dylan/tests/constants.dylan
+++ b/sources/dylan/tests/constants.dylan
@@ -8,7 +8,7 @@ Warranty:     Distributed WITHOUT WARRANTY OF ANY KIND
 
 // Helper function to force evaluation at run-time,
 // instead of compile-time.
-define function run-time (x)
+define not-inline function run-time (x)
   x
 end function run-time;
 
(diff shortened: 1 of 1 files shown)


sources/dylan/tests/constants.dylan
```diff
@@ -6,6 +6,12 @@ Copyright:    Original Code is Copyright (c) 1995-2004 Functional Objects, Inc.
 License:      See License.txt in this distribution for details.
 Warranty:     Distributed WITHOUT WARRANTY OF ANY KIND
 
+// Helper function to force evaluation at run-time,
+// instead of compile-time.
+define function run-time (x)
```
housel: Best to declare this `not-inline`

pedro-w: Done, let me know if I should squash.

PR #1233, opened 2019-04-25
system: Fix testworks-specs failures

cgay:
* Comment out some testworks specs that don't work yet.
* Add generic functions that specs expected to exist.

### DIFF
--- a/sources/system/tests/specification.dylan
+++ b/sources/system/tests/specification.dylan
@@ -20,11 +20,13 @@ define module-spec date ()
                         <integer>, <integer>)
            => (<date>);
   function decode-date (<date>)
-            => (<integer>, <integer>, <integer>, <integer>,
-                <integer>, <integer>, <day-of-week>, <integer>);
+           => (<integer>, <integer>, <integer>, <integer>,
+               <integer>, <integer>, <day-of-week>, <integer>);
   function parse-iso8601-string (<string>) => (<date>);
+/* Commented out until https://github.com/dylan-lang/testworks/issues/97 is fixed.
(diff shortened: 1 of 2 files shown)


sources/system/tests/specification.dylan
```diff
@@ -250,9 +252,10 @@ define module-spec simple-xml ()
   // Functions
   function document-location (<xml-document>) =>  (<locator>);
   function document-location-setter (<locator>, <xml-document>) =>  (<locator>);
-  function document-element (<xml-document>) =>  (<xml-element>);
-  function document-element-setter (<xml-element>, <xml-document>) =>  (<xml-element>);
+  function document-element (<xml-document>) => (false-or(<xml-element>));
+  function document-element-setter (false-or(<xml-element>), <xml-document>) => (false-or(<xml-element>));
   function read-xml-document (<locator>) =>  (<xml-document>);
+  // (A bit odd that these are defined on <xml-element> rather than <xml-node>.)
```
housel: Only elements in XML have attributes

cgay: Yes but the functions are named `node-*`, which seems misleading to me, especially given that those for `<xml-document>` are prefixed `document-` rather than `node-`.  I'll rephrase the comment to reflect that.

Come to think of it, shouldn't `<xml-node>` be the root of the hierarchy?:
```
<xml-node>
  <xml-element>
    <xml-document>
  [<xml-data>]
  [<xml-comment>]
```
I'm just going by https://www.w3.org/XML/Datamodel.html

(And this module would be ideal to import with `prefix: "xml/"` if "xml-" were removed from all the names and all the node-/document- prefixes were dropped.)

Issue #96, opened 2011-12-14
C-FFI docs: Index


waywardmonkeys:
Use the old HTML as a guide to add index markup.

PR #1243, opened 2019-08-05
Docs #96: C-FFI index 

fraya:
Index of C-FFI completed based on [web archive](http://web.archive.org/web/20080523233647/http://www.opendylan.org/documentation/opendylan/interop1/inte_152.htm#HEADING152-0).

### DIFF
--- a/documentation/library-reference/source/c-ffi/index.rst
+++ b/documentation/library-reference/source/c-ffi/index.rst
@@ -1,3 +1,7 @@
+.. index::
+   single: C-FFI library
+   single: libraries; C-FFI
+
 *****************
 The C-FFI Library
 *****************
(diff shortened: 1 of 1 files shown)


documentation/library-reference/source/c-ffi/index.rst
```diff
@@ -75,6 +98,11 @@ by the class ``<C-double>``, and ``unsigned long`` by the class
    Dylan names appearing in this document, is not binding and can safely be
    ignored.
 
+.. index::
+   single: <C-double*> class
```
cgay: Does the "*" here need to be escaped or does it show up okay in the index?  I only noticed it because GitHub's diff display gets confused by it.

fraya: It shows OK in the index.

PR #1248, opened 2019-08-13
[Doc] Intro to Dylan: DRM links

### LINKED ISSUE #97: Intro to Dylan: DRM links
This should have proper links to the DRM using the :drm: role.

fraya:
Fix #97

Link words present in `drmindex.py` with `:drm:` role.

### DIFF
--- a/documentation/intro-dylan/source/conditions.rst
+++ b/documentation/intro-dylan/source/conditions.rst
@@ -42,9 +42,9 @@ conditions for you to examine.
 Handlers
 ========
 
-A function :term:`establishes a handler` with the :drm:`let handler
-<let_handler>` statement. The handler remains in effect until the function
-exits. Other functions called by the first can establish new handlers. When the
+A function :term:`establishes a handler` with the :drm:`let handler`
+statement. The handler remains in effect until the function exits. Other
+functions called by the first can establish new handlers. When the
 :drm:`signal` function looks for a handler, it looks for the most recently
 established handler that fits the condition.
 
@@ -147,8 +147,8 @@ depending on the color of the sky.
     end block;
 
 Many programs need to dispose of resources or perform other cleanup work when
-exiting a block. The block may contain optional :drm:`afterwards <block>` and ``cleanup``
-clauses. Neither affects the block's return value. The :drm:`afterwards <block>` clause
+exiting a block. The block may contain optional :drm:`afterwards` and ``cleanup``
+clauses. Neither affects the block's return value. The :drm:`afterwards` clause
 executes if the block ends normally without using its nonlocal exit, and the
 ``cleanup`` clause executes when the block ends whether it ends normally or via
 nonlocal exit.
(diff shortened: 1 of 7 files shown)


documentation/intro-dylan/source/conditions.rst
```diff
@@ -147,8 +147,8 @@ depending on the color of the sky.
     end block;
 
 Many programs need to dispose of resources or perform other cleanup work when
-exiting a block. The block may contain optional ``afterwards`` and ``cleanup``
-clauses. Neither affects the block's return value. The ``afterwards`` clause
+exiting a block. The block may contain optional :drm:`afterwards <block>` and ``cleanup``
+clauses. Neither affects the block's return value. The :drm:`afterwards <block>` clause
```
cgay: Maybe we can just add an index entry for "afterwards" to the DRM instead?

documentation/intro-dylan/source/modules-libraries.rst
```diff
@@ -153,7 +154,7 @@ use a similar syntax:
 
     define open generic sample-function (o :: <object>) => ();
 
-A third form, ``define sealed domain``, partially
+A third form, :drm:`define sealed domain <define_sealed_domain>`, partially
```
cgay: I wonder if it would be too magic to change https://github.com/dylan-lang/sphinx-extensions/blob/master/sphinxcontrib/dylan/domain/drmindex.py#L274 to replaces spaces with underscores so we don't have to repeat like this.  Just an idea.

documentation/intro-dylan/source/conditions.rst
waywardmonkeys: I agree.

documentation/intro-dylan/source/modules-libraries.rst
waywardmonkeys: I agree here too,

documentation/intro-dylan/source/modules-libraries.rst
```diff
@@ -13,7 +13,8 @@ Simple Modules
 Modules import names (or bindings) from other modules and export names
 for use by other modules. The names that may be imported/exported are
 the module-level (also called "global") variables such as those created
-by ``define variable``, ``define class``, ``define generic``, etc.
+by :drm:`define variable <define_variable>`, :drm:`define class
+<define_class>`, :drm:`define generic <define_generic>`, etc.
```
waywardmonkeys: As @cgay points out ... this would be much improved by an improvement to `drmindex.py`.

documentation/intro-dylan/source/conditions.rst
fraya: `afterwards` is not in DRM index and the only reference is in the `Other Statement macros` with `block`. Where should `afterwards` point to?

documentation/intro-dylan/source/modules-libraries.rst
fraya: Ok, I take note.

fraya: This change will affect [drm_link](https://github.com/dylan-lang/sphinx-extensions/blob/321d4e77302b1fff1f25ea8fe32272dd60e628f6/sphinxcontrib/dylan/domain/dylandomain.py#L37) and the validation of invalid syntax in `:dylan:drm` roles. Is that ok?

documentation/intro-dylan/source/conditions.rst
cgay: That sounds right, yes.  Same as block but it's own entry.

waywardmonkeys:
You'll need to submit a PR for your change to the Sphinx extension ...

PR #1251, opened 2019-08-27
common-dylan: Add end argument to replace-subsequence!

abeaumont:
The replacement size should be the same as sequence size.
If this is not explicitely indicated the whole result suffix will be
discarded, with a huge performance penalty for large sequences,
as this operation is done in the inner loop.


waywardmonkeys:
Also, in your commit comment, `explicitely` -> `explicitly`

abeaumont:
Thanks! Changed the commit description (and fixed the typo).

Issue #1239, opened 2019-05-21
(2 ^ 62) - 1 returns -1 on LLVM back-end


cgay:
`format-out("%d", (2 ^ 61) - 1)` gets an overflow error, as expected.

```
$ _build/bin/hello
Integer overflow: The result of the last operation is too large for this integer representation.
Backtrace:
  invoke-debugger:internal:dylan##1 + 0x29
  default-handler:dylan:dylan##1 + 0x12
  default-last-handler:common-dylan-internals:common-dylan##0 + 0x2e6
  signal:dylan:dylan##0 + 0x120
  error:dylan:dylan##0 + 0xb1
  -:dylan:dylan##1 + 0x16
  _Init_hello__X_hello_for_system + 0x6a
  _init + 0x151
  __libc_start_main + 0xf1
  _init + 0x62
```

`format-out("%d", (2 ^ 62) - 1)` prints `-1`.

```
$ _build/bin/hello
-1
```

housel:
Fixed in PR#1256.

pedro-w:
There is a compiler error when using the C back end, which I think was introduced by the regression test for this bug.
```
Cc /opendylan/_build/Bootstrap.3/build/libraries-test-suite-app/../libraries-test-suite/../common-dylan-test-suite/regressions.o : /opendylan/_build/Bootstrap.3/build/libraries-test-suite-app/../libraries-test-suite/../common-dylan-test-suite/regressions.c
:931:22: error: integer literal is too large to be represented in any integer type
  T0 = (dylan_value) 0xFFFFFFFFFFFFFFFF8000000000000001L;
                     ^
/opendylan/_build/Bootstrap.3/build/libraries-test-suite-app/../libraries-test-suite/../common-dylan-test-suite/regressions.c:932:31: error: integer literal is too large to be represented in any integer type
  MV_SET_ELT(0, (dylan_value) 0xFFFFFFFFFFFFFFFF8000000000000001L);
                              ^
2 errors generated.
```

PR #1256, opened 2019-09-24
dfmc-llvm-back-end: Fix left shift overflow detection

housel:
This change fixes the previously-incorrect LLVM back-end
implementation of the primitive-machine-word-shift-left-signal-overflow primitive.

This fixes issue #1239.

### DIFF
--- a/sources/common-dylan/tests/regressions.dylan
+++ b/sources/common-dylan/tests/regressions.dylan
@@ -29,7 +29,20 @@ define test bug-5954 ()
               -17);
 end test bug-5954;
 
+define test bug-left-shift-signal-overflow ()
+  // See https://github.com/dylan-lang/opendylan/issues/1239
+  assert-no-errors(ash(1, $machine-word-size - 4));
+  assert-no-errors(ash(-1, $machine-word-size - 3));
+  assert-signals(<arithmetic-overflow-error>,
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
Did you mean for this to include the testworks channges? They seem like a separate concern.

housel:
Yes, all other things being equal I like to commit regression tests together with the corresponding bug fixes.

pedro-w:
I think @waywardmonkeys  is referring to the changes to the testworks submodule which you can see [here](https://github.com/dylan-lang/opendylan/pull/1256/files#diff-4aa70a95fdb1c4aaa2a834ac4cafbfa1) - are they related to the change in hand?

housel:
Oh, no they are not... updated.

sources/common-dylan/tests/regressions.dylan
```diff
@@ -29,7 +29,19 @@ define test bug-5954 ()
               -17);
 end test bug-5954;
 
+define test bug-github-1239 ()
```
cgay: Bike shedding: to me it would be better to name the test for what it does, e.g. test-shift-left-signals-overflow, and put the bug URL in a comment. Up to you though.

housel: Fair enough. Updated again.

Issue #1068, opened 2018-01-19
Serious warning when compiling table.dylan


pedro-w:
I've looked into the error that is shown when compiling any dylan project; it comes from compiling table.dylan in the dylan library itself

````
sources/dylan/table.dylan:873: Serious warning - Invalid type for argument new-key in call 
to entry-key-setter (new-key :: <object>, v :: <entry-keys>, index :: <integer>) => 
(new-key :: <object>):  <raw-pointer> supplied, <object> expected.
````

This is a warning and the generated code appears to function correctly.

The cause appears to be an optimization applied to pairs of primitive coercion functions, like `primitive-byte-character-as-raw` and `primitive-raw-as-byte-character`. If you apply a function to its inverse, then both calls can be removed. This is important for the low level stuff, I believe.
A further optimization occurs if a function with an inverse is applied to the result of an if-merge; in this case the function is pushed backwards into both arms so that it has a chance to be cancelled out. This might affect code like
````
let x = if (something?) a-as-b(...) else a-as-b(...) end;
let y = b-as-a(x); // b-as-a is pushed back and b-as-a(a-as-b(...) is eliminated.
````
However it might be that `x` is used later on. The compiler therefore inserts extra nodes to allow this, so it ends up like:
````
let y = if (something?) ... else ... end;
let x = a-as-b(y);
````
The code to do that is here:
https://github.com/dylan-lang/opendylan/blob/102d13bcbc80215e1fbf7b679bf36867c9733bf6/sources/dfmc/optimization/calls.dylan#L224-L288

The problem is that `primitive-cast-pointer-as-raw` and `primitive-cast-raw-as-pointer` are not exactly inverses because if you convert an object to pointer and back again, it loses its type information. Hence the warning; the compiler wants an `<object>` but it has been converted via raw to just a `<raw-pointer>`

So  a quick fix is to remove the primitive-cast-pointer-as-raw and primitive-cast-raw-as-pointer pair from consideration, i.e. delete these lines
https://github.com/dylan-lang/opendylan/blob/102d13bcbc80215e1fbf7b679bf36867c9733bf6/sources/dfmc/optimization/calls.dylan#L329-L330

A proper fix might be to track the converted type or just suppress the optimization if there are users of the intermediated value. That would probably take a compiler expert to do correctly, I wouldn't be confident myself.

waywardmonkeys:
This is fascinating ... I'm not sure how much would be pessimized by removing those lines!

[...]

PR #1258, opened 2019-09-28
dfmc-optimization: Prevent type loss from raw pointer casts

housel:
This compiler change provides a fix for issue #1068, addressing the issues identified by @pedro-w there.


waywardmonkeys:
Does this have much (any?) impact on the useful optimization of generated code?

housel:
The only change I could detect in the DFM output for a full build was that it allowed the previously-failed inlining of a call to `entry-key-setter` (due to the type error) to succeed.

PR #1262, opened 2019-10-02
Use hexadecimal floating point notation in the output of the C back end

pedro-w:
The current C back-end uses `float-to-string` to emit normal (decimal) floats into the C source files. Unfortunately `float-to-string` has limitations which means it cannot represent all floats faithfully. This PR changes the C back-end to output hexadecimal floats which can represent `<single-float>` and  `<double-float>` with 100% fidelity.
As part of the implementation, `decode-float` now handles subnormal numbers properly.
Finally, there is some code to work around the breaking change to `decode-double-float` introduced in 1fde59f3734d6147b2006c4a7a0d2678bbbea5ce. Without this, it is not possible to bootstrap from 2019.1 using the C back end.

### DIFF
--- a/sources/common-dylan/numerics.dylan
+++ b/sources/common-dylan/numerics.dylan
@@ -98,6 +98,27 @@ define method classify-float
   end case
 end;
 
+// Assume that the float is actually subnormal,
+// i.e. exponent is zero and significand is non-zero
+define function decode-single-subnormal
+    (float :: <single-float>)
+ => (significand :: <single-float>, exponent :: <integer>,
+     sign :: <single-float>);
+  let sign :: <single-float> = if (negative?(float)) -1.0s0 else 1.0s0 end;
+  let decoded :: <machine-word> = decode-single-float(float);
+  let significand-bits :: <machine-word> = %logand(decoded, #x7FFFFF);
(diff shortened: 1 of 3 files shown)


sources/common-dylan/numerics.dylan
```diff
@@ -240,6 +261,43 @@ define method classify-float
...
+     exponent :: <integer>,
+     sign :: <double-float>);
+  let sign :: <double-float> = if (negative?(float)) -1.0d0 else 1.0d0 end;
+  let (decoded-low :: <machine-word>, decoded-high :: <machine-word>)
+    = decode-double-float(float);
+  let significand-low-bits :: <machine-word> = decoded-low;
+  let significand-high-bits :: <machine-word> = %logand(decoded-high, #xFFFFF);
+  let shift :: <integer>
+    = if (zero?(significand-high-bits))
+	%count-high-zeros(significand-low-bits)
```
housel: Oops, looks like you need to untabify this

pedro-w: Annoying. Sorry.

housel:
Could the whitespace changes be folded into the other commits?

pedro-w:
WS changes now folded.

PR #1266, opened 2019-10-07
Update testworks submodule + add json submodule

cgay:
json needed for new json report format


housel:
Needs a registry entry for testworks-run as well.

PR #1290, opened 2019-12-27
system: Add error signaling for load-library

### LINKED ISSUE #1283: load-library should signal errors
`load-library` in the `system` library does not do any special handling of error conditions; if the underlying `dlopen` or `LoadLibrary` fails it will just return a `NULL` `<machine-word>` value. Signaling some sort of error condition would be more appropriate.

housel:
Fixes #1283


cgay:
https://opendylan.org/documentation/library-reference/system/operating-system.html#system:operating-system:load-library could use a word on errors being signaled.

PR #1296, opened 2019-12-31
Common dylan test suite fixes

### LINKED ISSUE #1295: Make test suites pass
I think it's important to have all the test suites passing (i.e., green), even if it means marking some of the long term problems as expected to fail.  Running the test suites (e.g., make check) and being faced with 500 failures and 30 crashes makes it extremely difficult to know whether one's changes had any effect. I want to be able to run the tests and know that if they're GREEN then I didn't break any of them.

This bug serves as a place to link commits that move toward this goal.

cgay:
#1295

### DIFF
--- a/sources/common-dylan/byte-vector.dylan
+++ b/sources/common-dylan/byte-vector.dylan
@@ -20,10 +20,13 @@ define function byte-vector-ref
 end function byte-vector-ref;
 
 define function byte-vector-ref-setter
-    (value, byte-vector :: <byte-vector>, index :: <integer>)
-  element(byte-vector, index) := value;
-end function byte-vector-ref-setter;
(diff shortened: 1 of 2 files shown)


sources/common-dylan/byte-vector.dylan
```diff
@@ -44,7 +47,9 @@ define sealed method byte-vector-fill
 end method;
 
 //---*** It would sure be nice to have low-level run-time support for this
-define open generic copy-bytes (dst, dst-start, src, src-start, n) => ();
+define open generic copy-bytes
+    (dst :: <sequence>, dst-start :: <integer>,
```
housel: This could be `<mutable-sequence>`

cgay: Done

Issue #1047, opened 2016-03-01
LLVM: <C-signed-int> return types on 64 bit platforms


waywardmonkeys:
If you have C-FFI definitions that have a `result res :: <C-signed-int>` on a 64 bit build, then you end up with an error while compiling from calls to `PRIMOP wrap-abstract-integer`

```
Error: Type {<llvm-integer-type> #x01D04360 i32} does not match {<llvm-integer-type> #x0406EF30 i64}
```

According to @housel:

> **housel** brucem: I guess that's expected (since `<raw-c-signed-int>` isn't the same size as `<raw-machine-word>`)
> **brucem** housel: hmm ... so is the right change to do something in the LLVM back-end or in C-FFI?
> **housel** The right thing is to add a cast operator primitive and have C-FFI generate that
> **brucem** housel: How painful would that be?
> **housel** medium painful I guess

housel:
Fixed in PR #1302.

PR #1302, opened 2020-01-12
C-FFI fixes and improvements

housel:
These commits make a number of fixes and improvements to the C-FFI library, its unit tests, and related compiler and run-time support.

This change fixes issue #1047


cgay:
Do you think this warrants a release note?

housel:
I have a backlog of release notes already; I'll make sure that this gets included in that.

PR #1318, opened 2020-04-22
[io tests] Add benchmark for string streams

cgay:
Add benchmark for string streams, plus an alternative string builder and benchmark to compare against.

### DIFF
--- a/sources/io/tests/streams-benchmarks.dylan
+++ b/sources/io/tests/streams-benchmarks.dylan
@@ -1,0 +1,88 @@
+Module: io-test-suite
+Synopsis: streams benchmarks
+
+// These first two benchmarks compare writing to a string stream and building a
+// string with a simple <string-builder> class, modeled after the one in Go.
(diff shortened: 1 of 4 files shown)


sources/io/tests/streams-benchmarks.dylan
```diff
@@ -0,0 +1,84 @@
...
+    init-keyword: buffer:;
+  slot builder-index :: <integer> = 0;
+end;
+
+define method builder-write (b :: <string-builder>, s :: <string>)
+  let idx = b.builder-index;
+  let s-size = s.size;
+  let new-size = idx + s-size;
+  if (new-size >= b.builder-buffer.size)
+    builder-grow(b);
```
cgay, line 48: Looking at this code afresh, I think `new-size` needs to be passed to `builder-grow`, which should make sure to create a new buffer bigger than that.

cgay:
Fixed the issue I mentioned above and added the benchmarks to the suite.

PR #1322, opened 2020-06-08
Registry cleanups

cgay:
These are motivated by wanting to make `dylan-tool update` work reliably in the OD repository, but they're good cleanups to have done in any case.

There's more to do, but it's mostly related to DUIM and Windows-only libraries that I can't currently test easily so I leave those to another time.


pedro-w:
Maybe related? the registry entries for `factorial-big` and `factorial-small` don't point to the right directories.
https://github.com/dylan-lang/opendylan/blob/2a12f0fb0eda6ad062a19c26658d8a351bf1913c/sources/registry/generic/factorial-big#L1
the sources actually are in `examples/console/factorial`

cgay:
> Maybe related? the registry entries for `factorial-big` and `factorial-small` don't point to the right directories.
> https://github.com/dylan-lang/opendylan/blob/2a12f0fb0eda6ad062a19c26658d8a351bf1913c/sources/registry/generic/factorial-big#L1
> 
> 
> the sources actually are in `examples/console/factorial`

I think not directly related. I suspect there are quite a few registry entries that are obsolete/broken, and that's one reason I want to auto-generate the registry. 

But you've reminded me to verify that none of the files I'm removing are referenced by existing registry files! Will do that.

cgay:
I checked `sources/registry` and didn't find any of the deleted files here to be referenced.

PR #1327, opened 2020-08-09
Locators improvements

cgay:
@housel would you be up for trying this out on Windows? I don't have a working Windows machine at the moment.

### DIFF
--- a/documentation/library-reference/source/io/print.rst
+++ b/documentation/library-reference/source/io/print.rst
@@ -20,7 +20,7 @@ of the overhead to satisfy special print requests, outputting any
 special notations required, and it only calls :gf:`print-object` when it
 is necessary to print objects. Users should always call the
 :func:`print` function to output objects, especially recursively from
-within :gf:`print-object` methods to output an object's components.
+within :gf:`print-object` methods, to output an object's components.
 Users should never call :gf:`print-object` directly.
 
 The IO library exports two modules for use with printing, *print* and
@@ -46,6 +46,7 @@ implementation-dependent.
 - :var:`*print-length*`
 - :var:`*print-circle?*`
 - :var:`*print-pretty?*`
+- :var:`*print-escape?*`
 
 Pretty printing
 ---------------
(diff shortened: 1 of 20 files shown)


sources/system/file-system/win32-file-system.dylan
```diff
@@ -65,6 +65,13 @@ define function %shorten-pathname
 end function %shorten-pathname;
 
 
+// TODO(cgay): implement this for Windows. is there something similar to realpath?
+define function %resolve-locator
+    (locator :: <microsoft-file-system-locator>)
+ => (resolved-locator :: <microsoft-file-system-locator>)
+  locator
```
housel, line 72: Here is a real implementation:
```dylan
define function %resolve-locator
    (locator :: <microsoft-file-system-locator>)
 => (resolved-locator :: <microsoft-file-system-locator>)
  with-stack-path (path-buffer)
    let path-length = raw-as-integer
                        (%call-c-function ("GetFullPathNameA", c-modifiers: "__stdcall")
                           (fileName :: <raw-c-pointer>,
                            bufferLength :: <raw-c-unsigned-long>,
                            bufferPtr :: <raw-c-pointer>,
                            filePart :: <raw-c-pointer>)
                           => (bufferUsed :: <raw-c-unsigned-long>)
                           (primitive-string-as-raw(as(<byte-string>, locator)),
                            integer-as-raw($MAX_PATH),
                            primitive-string-as-raw(path-buffer),
                            primitive-cast-raw-as-pointer(integer-as-raw(0)))
                        end);
    if (path-length = 0)
      if (win32-raw-last-error() = $ERROR_NOT_SUPPORTED)
        locator
      else
        win32-file-system-error("resolve", "%s", locator)
      end
    elseif (path-length > $MAX_PATH)
      win32-file-system-error("resolve", "%s", locator)
    elseif (~%file-exists?(locator))
      win32-file-system-error("resolve", "%s", locator)
    else
      as(object-class(locator), copy-sequence(path-buffer, end: path-length))
    end
  end
end function;
```

documentation/library-reference/source/io/print.rst
```diff
@@ -159,21 +162,62 @@ IO library's *print* module.
      ``#t`` ; however, when printing exits the dynamic scope of the call
      that changed the value to ``#t``, the value reverts back to ``#f``.
 
+     The *escape?* keyword indicates whether printing should show escape
+     codes in strings and, by metaphor, whether other objects should be
```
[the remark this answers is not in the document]
cgay: Done

sources/system/file-system/win32-file-system.dylan
cgay: I incorporated this, untested.

cgay:
> Some of the commit header lines exceed the Git recommended width

Done

PR #1334, opened 2020-09-13
common-dylan, io: simplify adding debug info to objects printed rep

cgay:
Export debug-name from common-dylan. This makes it easier to add identifying
information to objects for display during debugging, by simply adding

  `define method debug-name (o :: <my-object>) ... end`

without needing to find and import all of these definitions:
```dylan
  dylan:dylan-extensions:debug-name
  io:print:print-object
  io:pprint:printing-logical-block (optional, undocumented)
  io:streams:<stream>
```

### DIFF
--- a/documentation/library-reference/source/common-dylan/common-extensions.rst
+++ b/documentation/library-reference/source/common-dylan/common-extensions.rst
@@ -198,6 +198,34 @@ The extensions are:
      The *format-string* is a format string as defined on page 112 of
      the DRM.
 
+.. generic-function:: debug-name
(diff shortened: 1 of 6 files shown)


cgay:
ok, closing this. I would prefer to have zero imports in order to add debug info, but only importing `io:print` is sufficient. Having to import both `io:print` and `io:pprint` (and the requisite searching for what is needed) was just too much.

cgay:
hmm, i forgot about `io:streams`, sigh.

PR #1342, opened 2020-10-11
strings: document new find-any function

cgay:
This came about because I noticed that the cl-strings library duplicates much of what is now in the strings library. (More accurately, it was the other way around since the strings library came later.) I don't think we have anything quite like `string-search-set`, and that was the only thing besides `string-pluralize` that we don't have an equivalent for in the `strings` library.

An alternative would be to modify `common-dylan:common-extensions:position`:

```dylan
define open generic position
    (sequence :: <sequence>, target,
     #key test :: <function>,
          start: _start :: <integer>,
          end: _end,
          skip :: <integer>,
          //---*** For compatibility: remove this eventually                                                                                                                           
          count)
 => (position :: false-or(<integer>));
```

I want a way to search from the end, and I've never understood the point of the `skip` argument.
Thoughts?

### DIFF
--- a/documentation/library-reference/source/strings/index.rst
+++ b/documentation/library-reference/source/strings/index.rst
@@ -1180,6 +1180,35 @@ convenience.  (See `DEP-0004
 Miscellaneous Functions
 -----------------------
 
+.. generic-function:: find-any
+   :sealed:
+
+   Find the index of any character matching a predicate function.
+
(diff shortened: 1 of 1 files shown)


cgay:
The corresponding code was committed in https://github.com/dylan-lang/strings/commit/aad35123e4a145fa4a06c3b6ff062e3697d4097f

PR #1353, opened 2020-11-27
Upper/lowercase formatting of hex integers

### LINKED ISSUE #1054: Formatting hex numbers with either lowercase or uppercase A-F
Sometimes, it would be nice if we could display hex numbers using lowercase `a-f` values rather than `A-F`.

To do this, we'd want to handle `%x` and `%X` differently in `format`.

We should also make it possible for `integer-to-string` to produce either lower or uppercase.

(Changes to `format` should be postponed until one of my branches lands as it changes that code a fair bit.)

cgay:
Just trying to pick off an easy bug, #1054   3-stage bootstrap works. I didn't update all current uses of `%x` to be `%X` but that would be the minimal change. However, I feel like this is pretty safe since I believe C code can generally handle both upper and lowercase hex numbers, and that seems like the most likely non-carbon-based consumer of this kind of output.

Fixes #1054

### DIFF
--- a/documentation/library-reference/source/common-dylan/common-extensions.rst
+++ b/documentation/library-reference/source/common-dylan/common-extensions.rst
@@ -653,6 +653,7 @@ The extensions are:
    :parameter base: An instance of :drm:`<integer>` (default 10).
    :parameter size: An instance of :drm:`<integer>` (default 0).
    :parameter fill: An instance of :drm:`<character>` (default 0).
+   :parameter lowercase?: An instance of :drm:`<boolean>` (default ``#f``).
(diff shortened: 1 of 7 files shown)


pedro-w:
Should there be a test for this, in `sources/common-dylan/tests/functions.dylan` maybe?

cgay:
> Should there be a test for this, in `sources/common-dylan/tests/functions.dylan` maybe?

There is!  :-)

pedro-w:
🤦‍♂️

housel (approved):

PR #1355, opened 2020-11-29
Fix bug in dfmc-reader:skip-multi-line-comments

cgay:
3-stage-bootstrap and dfmc-reader-test-suite pass

### DIFF
--- a/documentation/release-notes/source/2021.1.rst
+++ b/documentation/release-notes/source/2021.1.rst
@@ -20,6 +20,10 @@ Compiler
 * Fixed a crash in the optimizer that ocurred when compiling certain
   C-FFI primitive functions.
 
+* Multi-line comments that contain comment delimiters within strings are now
+  parsed correctly. For example: ``/* "//h/a/" */`` Fixes `bug 1325
+  <https://github.com/dylan-lang/opendylan/issues/1325>`_.
+
 Run-time
 ========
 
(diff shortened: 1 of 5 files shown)


housel:
It seems to me that the lexer is doing what it's specified to do by the DRM (i.e., it's not a bug per se), and that this is changing the spec as well as the implementation. I worry that it may be opening a can of worms, but in any case perhaps a DEP would be in order.

cgay:
Ah, I see, because the spec is silent about what to do if `//` is inside of a string. I made the assumption that `/* */` should be able to comment out arbitrary code, but as you say, that's only an assumption. ok, I'll write a DEP.

Can you be more specific about the worms? Just a generalized worry about delimited comments having to do more complex parsing? I couldn't think of anything other than strings that would cause a problem.

housel:
Another worm I thought of, for instance what should happen if you comment out `dylan/*` ?

cgay:
ok, fair point. It wouldn't be hard to make `skip-multi-line-comment` ensure that `/*` and `//` are not adjacent to an identifier, but how big is this can of worms...

I feel pretty strongly that one should be able to comment out arbitrary sequences of complete fragments and expect the compiler to act as if there's nothing there, so I'm inclined to push ahead and see if we can fix these cases. I've always found, for example, that in languages that don't allow nested `/* */` comments they are basically useless. I'm more likely to just use line comments on large regions of code, "just in case". (We're already in better shape than that, thankfully.)

PR #1361, opened 2020-12-14
common-dylan: fix bug in copy-bytes(<vector>, <string>)

### LINKED ISSUE #1360: write(<sequence-stream>, "ABC") writes integers not chars
https://play.opendylan.org/shared/de44270864d5003c
```
    let v = make(<vector>, size: 5);
    let stream = make(<sequence-stream>,
                      contents: v,
                      direction: #"output");
    write(stream, #(1, 2, 3));
    write(stream, "ABC");
    format-out("%=\n%=\n", v, stream-contents(stream));
```
results in
```
#[1, 2, 3, #f, #f]
#[1, 2, 3, 65, 66, 67]
```
but I would expect to see
```
#[1, 2, 3, #f, #f]
#[1, 2, 3, 'A', 'B', 'C']
```

cgay:
`write(<sequence-stream>, <string>)` ends up calling `copy-bytes(<vector>,
<string>)` when the underlying sequence is a `<vector>`.  `copy-bytes` isn't
documented, but I infer from the implementations that it is misnamed and should
be called `copy-elements` or `copy-sequence-into!`, and my fix makes that
assumption. That is, that it isn't just intended to copy bytes.

```
Before:
test-bug-1360 failed
  #[1, 2, 3, 'A', 'B', 'C'] = stream-contents(stream) failed [#[1, 2, 3, 'A', 'B', 'C'] and #[1, 2, 3, 65, 66, 67] are not =.  element 3 is the first non-matching element]

After:
test-bug-1360 passed in 0.000216 seconds with 26016 bytes allocated.
  #[1, 2, 3, 'A', 'B', 'C'] = stream-contents(stream) passed
```

Fixes #1360

### DIFF
--- a/sources/common-dylan/byte-vector.dylan
+++ b/sources/common-dylan/byte-vector.dylan
@@ -92,7 +92,7 @@ define open method copy-bytes
      src :: <string>, src-start :: <integer>, n :: <integer>)
  => ()
   for (i :: <integer> from 0 below n)
-    dst[dst-start + i] := as(<integer>, src[src-start + i])
+    dst[dst-start + i] := src[src-start + i]
   end
 end method;
 
--- a/sources/io/tests/streams.dylan
+++ b/sources/io/tests/streams.dylan
@@ -789,7 +789,7 @@ define test test-stretchy-stream (description: "<string-stream> stretchy vector 
           as(<stretchy-vector>, list(1,2,3,4,5,6))  );
   end;
 
- begin
+  begin
     let v = make(<vector>, size: 3);
     let s = make(<string-stream>, contents: v, direction: #"output");
     write(s, #(1, 2, 3));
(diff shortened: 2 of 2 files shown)


sources/common-dylan/byte-vector.dylan
```diff
@@ -92,7 +92,7 @@ define open method copy-bytes
      src :: <string>, src-start :: <integer>, n :: <integer>)
  => ()
   for (i :: <integer> from 0 below n)
-    dst[dst-start + i] := as(<integer>, src[src-start + i])
+    dst[dst-start + i] := src[src-start + i]
```
housel, line 95: Perhaps there shouldn't be methods on general `<vector>` at all. (We can leave this for now, but for the unicode branch I'll be seriously curtailing the use of `copy-bytes` where it's not appropriate.)

housel (approved):

PR #1370, opened 2021-01-05
Common dylan test suite

cgay:
```
testworks-run PASSED in 88.892943 seconds:
  Ran 30 suites: 30 passed (100.00000%)
  Ran 473 tests: 318 passed (67.230448%), 0 failed, 0 skipped, 155 not implemented, 0 crashed
  Ran 1 benchmark: 1 passed (100.00000%)
  Ran 2308 checks: 2306 passed (99.913344%), 2 failed, 0 skipped, 0 not implemented, 0 crashed
```

(The 2 failed checks are actually expected failures.)


sources/common-dylan/tests/specification.dylan
```diff
@@ -486,6 +486,77 @@ define interface-specification-suite threads-specification-suite ()
...
+  test test-automatic-finalization-enabled?;
+  test test-integer-length;
+  test test-float-radix;
+  test test-float-digits;
+  test test-float-precision;
+  test test-decode-float;
+  test test-scale-float;
+  test test-classify-float;
+end;
+
```
cgay, line 566: I probably should have reorganized these into a numerics suite, random suite, etc, but my plan is to delete the suites so I would rather just leave as is.

housel: I frequently use the existing suites and would prefer that they not be deleted.

sources/common-dylan/tests/classes.dylan
```diff
@@ -6,9 +6,10 @@ Copyright:    Original Code is Copyright (c) 1995-2004 Functional Objects, Inc.
 License:      See License.txt in this distribution for details.
 Warranty:     Distributed WITHOUT WARRANTY OF ANY KIND
 
-/// Class test cases
-
-// Note: Locators are tested in system-test-suite.
+// TODO(cgay): Everything in this file tests things defined in dylan-extensions
+// and should be tested in a dylan-extensions-test-suite, not here, where they
```
housel, line 10: The fact that things are in dylan-extensions is normally an accident of the implementation, i.e. they are defined in the dylan library for sealing, run-time integration or other reasons, but with the intention of being re-exported from common-dylan or elsewhere. Making dylan-extensions too much of a "thing" unnecessarily exposes this sort of implementation detail.

cgay:
Part of #1295

sources/common-dylan/tests/specification.dylan
cgay: How do you use them in practice?  It seems unlikely that you would want to, say, run the tests for all the functions or run the tests for all the macros, and that's the current organization. My usual M.O. is to run the whole thing once and then run a single failing test until I make it pass.

At any rate, I'll re-org the suites instead of deleting them for now. I'll continue to try and come up with a better design than suites because I still believe https://gitter.im/dylan-lang/general?at=5cb4d72d016a930a45c58c8e. I think I have something reasonable in mind.

cgay: Maybe I can't think of anything better after all. Best I can come up with is `define [suite-name] test test-name ()`.

sources/common-dylan/tests/classes.dylan
cgay: Fixed. PTAL (Please take another look)

housel (approved):

PR #1387, opened 2021-04-17
cl-strings: delete everything except string-pluralize

cgay:
Let's just have one strings library.  The one function that is really unique here is `string-pluralize`, so I left that.

Tested with 3-stage bootstrap and `make check`.


housel:
`string-pluralize` itself should go away as soon as we have a real internationalization system that can handle plurals in languages other than English.

housel (approved):

Issue #1269, opened 2019-10-08
Need a way to organize multiple documentation "projects"


cgay:
We need a good way to provide links between the Library Reference, Hacker Guide, Getting Started, and others, each of which is a separate Sphinx "project".

[InterSphinx](http://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) might be a good way to do this.

Long term, I picture the Library Reference being broken apart into individual projects for each library, and documenting those alongside documentation from all projects in the [pacman catalog](https://github.com/cgay/pacman-catalog).

cgay:
I posted https://groups.google.com/g/sphinx-users/c/CcnnXV2LOoY but got no response.

cgay:
The answer to my question turned out to be pretty darned simple but the intersphinx docs don't make it very clear.  If we have 
```python
intersphinx_mappings = {
    'http': ('https://opendylan.org/documentation/http/', None)
}
```
Then a link to the `add-resource` function in the HTTP docs looks like this
```
:func:`http:add-resource`
```

And there was much rejoicing.

cgay:
See the above commit for a start, but there are some issues remaining to work out, as mentioned in that PR. Also on sphinx-users they said (if I understood correctly) that both of these should work, barring multiple definitions with the same name in different libraries:
```
:func:`library-reference:foo`
:func:`foo`
```
So that needs testing.

cgay:
Also also, we need to be able to test the `intersphinx_mapping` locally before publishing doc changes. There's a [mechanism for this](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html?highlight=intersphinx#confval-intersphinx_mapping) but I decided to leave it for a later time.

cgay:
This has all been fixed by the recent website revamp. Everything is built as a unit and references are just Sphinx `:doc:...` references.

PR #1393, opened 2021-05-02
doc: add intersphinx_mapping to Library Reference

cgay:
Enables linking between independently built documents like the Hacker Guide and the Testworks docs with syntax like this:

```
:func:`library-reference:*standard-output*`
```

I added a couple of such references in this commit as proof of concept. There are some mysteries remaining to be solved, such as why this doesn't work:

```
:class:`library-reference:<bit-set>`
```

Rather than repeat the `intersphinx_mapping` configuration in each document I factored it out into `shared_sphinx_config.py` and load that with `... import *` in each document's `conf.py`, along with the `extensions` setting. (More configs could be moved into the shared config file over time.)

### DIFF
--- a/documentation/building-with-duim/source/conf.py
+++ b/documentation/building-with-duim/source/conf.py
@@ -13,21 +13,19 @@
 
 import sys, os
 
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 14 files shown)


housel (approved):

cgay:
https://github.com/dylan-lang/opendylan/issues/1269

PR #1398, opened 2021-08-20
Update TLS support

phongphan:
Hi,

A few weeks ago I was trying to use Dylan for a 3rd party api validation program by hacking the `http` module for https support [here](https://github.com/phongphan/http/commit/ac2a98bc5994d6e8b16cb87cde239c6533faaa9b).

But since the target is behind some sort of the CDN, I have to update the  `network` and `ssl` packages as described below. 
* Update ssl module for openssl@1.1 or libressl
* Add SNI support to the client ssl
* Patch the network module to always return the `initialization-name` for `new-address.%hostname`.
  * The `gethostbyname` on macOS seems to resolve the h-name to CDN host name. Eg. `www.microsoft.com` was resolved to `e13678.dscb.akamaiedge.net` which is not usable for the SNI above.

At least I hope it would be useful for the future update. Or any suggestion that would convert them to a proper patch would be appreciated :)

### DIFF
--- a/sources/lib/ssl/c-wrapper.dylan
+++ b/sources/lib/ssl/c-wrapper.dylan
@@ -103,49 +103,19 @@ end;
 //hope that I can treat this as opaque
 define constant <SSL-METHOD> = <C-void*>;
 
-define C-function SSLv2-method
+define C-function TLS-method
   result ssl-method :: <SSL-METHOD>;
-  c-name: "SSLv2_method"
+  c-name: "TLS_method"
 end;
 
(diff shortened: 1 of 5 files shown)


cgay:
I won't have time to look at this this weekend but will try to look at it on Monday.

phongphan:
Hi, I have made an update to remove the `SSL-ctrl` from the history.

housel (approved):

cgay (approved):

cgay:
Thanks for the contribution!

PR #1399, opened 2021-10-03
Method Entry Point and other miscellaneous fixes

housel:


sources/dfmc/llvm-back-end/llvm-entry-points.dylan
```diff
@@ -638,15 +638,93 @@ end entry-point-descriptor;
...
+        let optionals-size
+          = call-primitive(be, primitive-vector-size-descriptor,
+                           optionals-cast);
+        let cmp = ins--icmp-slt(be, optionals-size, -count);
+        ins--if (be, op--unlikely(be, cmp))
+          op--argument-count-error(be, meth, optionals-size);
+        end ins--if;
+
+        let new-optionals-size = ins--add(be, count, optionals-size);
+      let new-optionals
```
cgay, line 667: Looks like this line through `end method` need to be indented two more spaces.

cgay (approved):

PR #1404, opened 2021-11-25
printing-object macro: Include object address

cgay:


sources/io/print.dylan
```diff
@@ -1326,18 +1326,41 @@ define method do-printing-object
...
     end
   end
 end method do-printing-object;
+
+// Cache of integer IDs we use instead of the object address when printing, in case we're
+// using a moving GC. <object> => <integer>
+define constant $object-id-cache :: <table> = make(<table>, weak: #"key");
+
+define variable *next-object-id* :: <integer> = 1;
+define constant $next-object-id-lock :: <lock> = make(<lock>);
```
housel, line 1350: Instead of a lock, could use `atomic-increment!`

cgay: Yep, I considered that but couldn't see a way to make it work.

housel (approved):

Issue #1389, opened 2021-04-26
file-system docs missing some exported functions


cgay:
These functions aren't documented:
* link-target
* directory-contents

Review the export list and document what's not documented.

PR #1409, opened 2021-12-08
file-system: Document link-target and directory-contents

cgay:
Just enough documentation that people can find these functions.
This addresses #1389, but I didn't review the entire export list
to see what else is undocumented.

### DIFF
--- a/documentation/library-reference/source/system/file-system.rst
+++ b/documentation/library-reference/source/system/file-system.rst
@@ -44,6 +44,7 @@ perform file management tasks at any position in the file system.
 
 - :func:`create-directory`
 - :func:`delete-directory`
+- :func:`directory-contents`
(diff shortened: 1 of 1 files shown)


documentation/library-reference/source/system/file-system.rst
```diff
@@ -186,6 +188,20 @@ File-System module.
      refers to a link, the link is removed, but the actual file that the
      link points to is not removed.
 
+.. function:: directory-contents
+
+   Returns a list of files and subdirectories contained in a directory.
```
housel, line 193: Returns a _sequence_ of ...

cgay: Thanks. I also added that `.` and `..` are not included in the result, for the avoidance of doubt.

PR #1410, opened 2021-12-22
DFMC Improved source-location tracking

housel:
These commits greatly reduce the number of DFM computations without source location info (identified with @??? in the DFM dump files), and thereby improve the DWARF debug info output by the LLVM back-end (as well as the CV4 debug info output by the HARP back-end).


cgay:
Sounds amazing. Definitely worthy of a release note, whether you want to do it here or in a separate PR.

PR #1413, opened 2022-01-03
[locators] Return "." if locator-name is #"self"

cgay:
`#"self"` doesn't match the return type of the function.


sources/system/file-system/posix-locators.dylan
```diff
@@ -69,7 +69,12 @@ define sealed method locator-name
  => (name :: false-or(<string>))
   let path = locator.locator-path;
   unless (empty?(path))
-    path[size(path) - 1]
+    let name = path[size(path) - 1];
+    if (name = #"self")
```
housel, line 73: Use `==` for symbols

cgay: Done

PR #1427, opened 2022-03-20
Miscellaneous fixes

housel:


cgay (commented):
For a change like this I would prefer the commit comment to say more about why the change is necessary or what the effect of the change is.

cgay (approved):

Issue #1373, opened 2021-01-07
Make dylan-compiler exit status reflect warnings


cgay:
It would be useful in various automation tasks if the `dylan-compiler` command would return non-zero exit status when serious warnings are encountered.

```shell
$ dylan-compiler -build testworks-run
.../testworks/command-line.dylan:23.20-38: Serious warning - Reference to undefined binding {whatever in %testworks}.
...
Build of 'testworks-run' completed

$ echo $?
0
```

Currently it's very easy to miss serious warnings in amongst all the other compiler output.

What issues would this raise, if any, for our current processes?  Should non-serious warnings be reflected in the exit status as well?

cgay:
Dunno if it's overly complex, but we could use the low three bits of the exit status to indicate
1. serious warnings
2. non-serious warnings
3. other errors

pedro-w:
I usually just ignore non-serious warnings in the first instance, e.g there are some when compiling the standard library, and also 'binding defined but not referenced or exported' happens all the time during development. But I find serious warnings mean that the program will crash. So it would definitely be useful to change the exit status on serious warnings, then I could do `dylan-compiler -build myprog && ./_build/bin/myprog` - as you say, if there's a lot of compiler output, a serious warning can be missed.
Maybe the compiler needs the equivalent of `-Werror` for when we want non-serious warnings to affect the exit status.

cgay:
And since Open Dylan has the property that even serious warnings don't prevent a binary being created, we should have to explicitly ask for the exit status to reflect that as a failure.  (Or at least there should be a way to turn it off.)

There's also the question of whether any flag to turn on/off this behavior would apply to the entire build or would mean to stop after compiling any library dependency that generates warnings, or only after the entire build is complete.  The former would be good for CI, to reduce overall resource usage. The latter is better for development, or at least is more like current behavior.

Straw man proposal:

1. For now any options apply to the whole build. Later if we want an early exit for warnings generated by library deps we can add a separate flag for that.
2. `--error-status-on-warnings={all,serious,none}`
3. Default to none, same as today. (This is conservative. I'd be fine with defaulting to serious.)

cgay:
Ended up returning non-zero exit code for serious warnings by default and providing `-allow-serious-warnings` to disable that behavior.

PR #1439, opened 2022-08-08
Console compiler: exit with error status for serious warnings

cgay:
Console compiler: exit with error status for serious warnings

Issue #1373

### DIFF
--- a/sources/environment/commands/command-line.dylan
+++ b/sources/environment/commands/command-line.dylan
@@ -8,8 +8,13 @@ Warranty:  Distributed WITHOUT WARRANTY OF ANY KIND
 
 /// Useful constants
 
-define constant $success-exit-code = 0;
-define constant $error-exit-code   = -1;
+// These are sorted by severity so that Windows scripts can use IF ERRORLEVEL.
+// https://devblogs.microsoft.com/oldnewthing/20080926-00/?p=20743
+define constant $success-exit-code          = 0;
+define constant $usage-error-exit-code      = 1;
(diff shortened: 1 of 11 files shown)


housel:
Why not add a command-line option to suppress this new exit status behavior, which could then be used when building libraries that deliberately induce serious warnings (such as the common-dylan-test-suite)?

cgay:
I was planning to add that flag in a later commit (please comment on the bug if you have preferences for what it should look like). 

I would like to be able to detect _new_ serious warnings in the test suite and Open Dylan code, which is why I prefer to comment out those two tests. I could make the exit code be the total number of serious warnings (maxing out at 127, I guess, if that's a shell limitation...127 should be enough serious warnings for anyone) and then callers who expect non-zero could compare with that?

sources/environment/commands/command-line.dylan
```diff
@@ -8,6 +8,11 @@ Warranty:  Distributed WITHOUT WARRANTY OF ANY KIND
 
 /// Useful constants
 
+define constant $success-exit-code          = 0;
+define constant $unexpected-error-exit-code = 1;
+define constant $usage-error-exit-code      = 2;
```
housel, line 13: Maybe sort these by seriousness so they could be used with `IF ERRORLEVEL` https://devblogs.microsoft.com/oldnewthing/20080926-00/?p=20743

housel (approved):

sources/environment/commands/command-line.dylan
cgay: Done

Issue #1354, opened 2020-11-28
log(0.0) infinite loop


cgay:
OD 2020.1, x86_64-linux, llvm

`log(0.0)` goes into an infinite loop.

cgay:
Note that sometimes the easiest thing may be to use the playground: https://play.opendylan.org/shared/7b5a2926995b709a

pedro-w:
> If x <= 0 <= 1, an error is signalled.

That '<= 1' doesn't make sense to me.

Another data point for you: on MacOS (10.15 64-bit,) `log(0.0)` gives `-{infinity}`, as does `log(-0.0)` and `log(-1.0)` gives
```
LOG would produce a complex number
Breaking into debugger.
Backtrace:
  0
```

housel:
See additional background at https://gitter.im/dylan-lang/general?at=632a097611a6a83d04c448cd

cgay:
The above link doesn't work for me (anymore). This link seems to work for me, to find where the discussion started: https://app.element.io/#/room/#dylan-lang_general:gitter.im/$ersG93YXlTKMgFYkA62O-Jg5eqTY4uxyOQDmbKDsSDs

PR #1449, opened 2022-09-16
[doc] Fix bugs in log and logn docs

cgay:
Discussed in #1354


cgay:
I could obsess for hours as to what the "right" way to format these things is. I notice that other places only use code quotes on the variables and not on the operator. e.g. `x` / `y` instead of `x / y`.  I'm going to expand this PR to do the latter everywhere.

This is made worse by the hideous red we use for code, but that's a problem for another day.

PR #1454, opened 2022-10-28
test-hypot: no longer expected to fail on macOS

cgay:
No idea when or how this was fixed, but it passes now.

#1295


cgay:
Need to investigate more. Closing for now.

PR #1477, opened 2023-02-05
Intro Dylan doc: fix example and missing :term:s

cgay:
This fixes the bug reported by "nsxfh"

```diff
-        make(<truck>, keys, tons: 2)
+        apply(make, <truck>, tons: 2, keys)
```

as well as many missing `:term:` targets. The latter were mostly references to the DRM, which have been created in
https://github.com/dylan-lang/sphinx-extensions/pull/26. For a few I just removed the `:term:` markup.

### DIFF
--- a/documentation/intro-dylan/source/methods-generic-functions.rst
+++ b/documentation/intro-dylan/source/methods-generic-functions.rst
@@ -2,7 +2,7 @@
 Methods & Generic Functions
 ***************************
 
-Dylan :term:`methods` correspond roughly to the
+Dylan :drm:`methods` correspond roughly to the
 functions found in C++. They take zero or more named parameters,
 but also return zero or more named return values. A minimal Dylan method
 might look like the following:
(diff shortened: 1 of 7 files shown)


housel (approved):

documentation/intro-dylan/source/methods-generic-functions.rst
```diff
@@ -48,7 +48,7 @@ return value (or actually does anything). As in C, each typed parameter
 must have its own type declaration; there's no syntax for saying
 "the last three parameters are all integers".
 
-Functions with variable numbers of parameters include the
+Functions with a variable numbers of parameters include the
```
housel, line 51: a variable number of

cgay: Actually I changed it to "Functions with the `#rest` keyword in their parameter list accept a variable number of arguments."  I think this gets causality right, and uses "arguments" rather than "parameters" for those who still believe in that pedantic distinction, like myself. :)

PR #1480, opened 2023-03-12
mps: Use MVFF pool, not MV.

waywardmonkeys:
The MV pool was deprecated long ago and removed in favor of using MVFF (first fit). The MVFF pool was in 1.108 which we still use on Windows, so it is safe to use here. This is required when updating to newer versions of MPS.

cc: @rptb1


rptb1 (approved):
[MVFF](https://memory-pool-system.readthedocs.io/en/latest/pool/mvff.html#pool-mvff) is a drop-in upgrade to MV with better scaling, especially to large heaps on 64-bit.

PR #1486, opened 2023-04-02
dfmc-reader: fix bug in make-hash-literal

cgay:
`#:foo:{\}` would crash the compiler with "element outside of range" unless the `}` delimiter occurred later in the same source record.

### DIFF
--- a/sources/dfmc/reader/interface.dylan
+++ b/sources/dfmc/reader/interface.dylan
@@ -109,6 +109,11 @@ define serious-program-warning <ratios-not-supported> (<invalid-token>)
   format-arguments token-string;
 end serious-program-warning;
 
+define serious-program-warning <unterminated-parser-expansion>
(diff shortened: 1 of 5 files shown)


sources/dfmc/reader/interface.dylan
```diff
@@ -109,6 +109,11 @@ define serious-program-warning <ratios-not-supported> (<invalid-token>)
   format-arguments token-string;
 end serious-program-warning;
 
+define serious-program-warning <unterminated-hash-literal>
+  format-string "Unterminated hash literal %s";
```
housel, line 113: "hash literal" is vague terminology and doesn't match the documentation.

housel (approved):

sources/dfmc/reader/interface.dylan
cgay: Done

PR #1509, opened 2023-06-28
library-reference: various fixes to numbers.rst

cgay:
* don't assume double integers are 64-bits
* link more names, especially DRM names
* some RST fixes

There's a **lot** more to do here.

### DIFF
--- a/documentation/library-reference/source/language-extensions/numbers.rst
+++ b/documentation/library-reference/source/language-extensions/numbers.rst
@@ -783,7 +783,7 @@ The generic-arithmetic Library
 ==============================
 
 The Generic-Arithmetic library exports the functions described in this
-section from an exported module called *generic-arithmetic*.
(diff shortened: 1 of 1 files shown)


documentation/library-reference/source/language-extensions/numbers.rst
```diff
@@ -1075,7 +1077,7 @@ example:
 
 .. code-block:: dylan
 
-    ga/+ (5, 4);
+    ga/+(5, 4);
```
housel, line 1080: In our codebase `generic/`, `generic-`, or `big/` is used; it might make sense to stick with one of these.

housel (approved):

documentation/library-reference/source/language-extensions/numbers.rst
cgay: done

PR #1510, opened 2023-06-29
Docs: Edit 'file-system' module

fraya:
- Add graphviz extension to create a class hierarchy to file system, posix and microsoft locators.

- Remove reference to 'file-stream' module. This paragraph was moved from module 'locators' and contained a reference to the module 'file-system', not needed now.

- Rename key parameter 'filename'. The key parameter appears as 'filename' and in the documentation of the method as 'locator'.

- To eliminate document duplication, I decided to address the issue in the paragraph discussing the 'make' method. Since it bears close resemblance to the method reference, I opted to remove the initial paragraph and provide a link to the reference instead. Additionally, I consolidated the documentation for the key parameters 'direction' and 'buffer-size' with a more comprehensive explanation.

### DIFF
--- a/documentation/library-reference/source/system/file-system.rst
+++ b/documentation/library-reference/source/system/file-system.rst
@@ -952,10 +952,8 @@ File-System module.
      instantiated, this method calls the generic function
      :gf:`type-for-file-stream`.
 
-     The *locator* init-keyword should be a string naming a file. If
-     the :doc:`Locators <../system/locators>` library is in use,
-     *locator* should be an instance of :class:`<locator>` or a string
-     that can be coerced to one.
+     The *locator* init-keyword should be a :class:`<file-locator>` or
+     a :drm:`<string>` that can be coerced to one.
 
      The *direction* init-keyword specifies the direction of the
      stream.  This can be one of ``#"input"``, ``#"output"``, or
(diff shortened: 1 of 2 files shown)


documentation/library-reference/source/system/file-system.rst
```diff
@@ -885,19 +952,22 @@ File-System module.
      instantiated, this method calls the generic function
      :gf:`type-for-file-stream`.
 
-     The *filename* init-keyword should be a string naming a file. If
+     The *locator* init-keyword should be a string naming a file. If
      the :doc:`Locators <../system/locators>` library is in use,
```
cgay, line 956: pre-existing: "library" here should be "module"

cgay (approved):
LGTM. Don't let me forget to install the graphviz extension on opendylan.org!

documentation/library-reference/source/system/file-system.rst
cgay: Actually I don't think we need to conditionalize this at all. Why not just say "The *locator* argument should be a ```:class:`<file-locator>` ``` or a ```:drm:`<string>` ``` that can be coerced to one.

cgay (approved):

PR #1517, opened 2023-07-05
Remove firing action on pull request

fraya:
- Action fails on pull request not having permissions. Delete it until we have more information.

- [doc] Fix error 'Explicit markup ends without a blank line; unexpected unindent'. Added here to test the action.


cgay:
Can we just run `ammaraskar/sphinx-action@master` for pull requests, so we get error checking, but then _also run_ `JamesIves/github-pages-deploy-action@v4` on push? I guess it could be done in two separate workflow files if it can't be done in one.

PR #1534, opened 2023-09-24
Implement DEP 12, string literals supporting Rectangle Rule

cgay:
See https://opendylan.org/proposals/dep-0012-string-literals.html#the-rectangle-rule for details.

This implements everything except for the ability to use three **or more** double quotes to start/end the multi-line string. Only exactly `"""` may be used at this point.

### DIFF
--- a/sources/dfmc/reader/lexer.dylan
+++ b/sources/dfmc/reader/lexer.dylan
@@ -861,8 +861,9 @@ end method hex-escape-character;
 define method decode-string
     (source-location :: <lexer-source-location>, bpos :: <integer>,
      epos :: <integer>, escapes? :: <boolean>)
- => (string :: <byte-string>)
+ => (string :: <byte-string>, multi-line? :: <boolean>)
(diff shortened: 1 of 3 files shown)


sources/dfmc/reader/lexer.dylan
```diff
@@ -922,9 +925,56 @@ define method decode-string
...
-  string
+  values(string, multi-line?)
 end method decode-string;
 
+// https://opendylan.org/proposals/dep-0012-string-literals.html#the-rectangle-rule
+//
+// When this is called `string` is known to contain at least one literal newline
+// character, the EOL sequence has already been canonicalized to just '\n', escape
+// sequences have been processed, and the start/end delimiters have been removed.
+define function trim-multi-line-prefix
```
housel, line 936: There are probably efficiency improvements that could be made to this function, but I guess it won't be a bottleneck in compiling most programs so it should be fine.

housel (approved):

PR #1560, opened 2024-02-20
Several simple doc fixes

cgay:


BUILDING.rst
```diff
@@ -47,24 +47,24 @@ x86 platforms.
...
 
-    brew install autoconf automake bdw-gc --universal
+    brew install autoconf automake bdw-gc
+
+.. note:: Open Dylan doesn't yet support native code on Apple silicon and
+          therefore ``bdw-gc`` should be installed for the ``x86_64``
+          architecture. One way to accomplish this is to install an ``x86_64``
+          version of Homebrew with ``arch -x86_64 /bin/bash -c "$(curl -fsSL
+          https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"``
+          and install with that.
```
waywardmonkeys, line 57: I don't think this is a good recommendation. You're suggesting something that would alter their entire install of homebrew, right?

BUILDING.rst
```diff
@@ -76,7 +76,7 @@ To build :program:`dylan-compiler` and several tools::
   ./autogen.sh
   ./configure --prefix=/opt/opendylan-current   # (but see note below)
   make
-  sudo make install
+  make install
```
waywardmonkeys, line 79: This is likely not correct in the general case.

waywardmonkeys (changes requested):

waywardmonkeys:
The first commit is going to need workshopping, but the other three are fine ... if you removed the first one from this PR, it'd be an easy merge to get that out of the way.

BUILDING.rst
cgay: ok, I dropped the brew-related changes and the sudo change but kept the changes to make version numbers less specific (so we don't have to keep updating them) and to point to release-with-batters.sh instead.

I guess the right way to install the GC is just to build it locally, a la release-with-batteries.sh?

waywardmonkeys (approved):
LGTM. The brew stuff should get updated in a subsequent PR as it still talks about x86 and `--universal`, but not worth doing that in this PR.

PR #1567, opened 2024-03-01
Rename $machine-name to $machine-architecture

cgay:
Keep old name for backward compatibility.  The uses in djam.dylan and jam-build.dylan were the only ones that couldn't be changed right away due to bootstrap requirements.

### DIFF
--- a/documentation/library-reference/source/system/operating-system.rst
+++ b/documentation/library-reference/source/system/operating-system.rst
@@ -20,7 +20,7 @@ environment variable in the system.
 The following constants contain machine-specific information:
 
 - :const:`$architecture-little-endian?`
-- :const:`$machine-name`
+- :const:`$machine-architecture`
 - :const:`$os-name`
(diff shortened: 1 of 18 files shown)


waywardmonkeys:
#1566 is probably going to conflict with this, so I'll wait for that to land and conflicts resolved before I review.

waywardmonkeys (approved):
I didn't check things outside of this PR like "Did you get every instance?" ... but what's in the PR looks good so assuming you did, then all is well.

housel:
`djam.dylan` isn't a bootstrapping dependency (since it's a standalone application not used during the bootstrapping process).

cgay:
> `djam.dylan` isn't a bootstrapping dependency (since it's a standalone application not used during the bootstrapping process).

Possibly it was just jam-build.dylan and I made the assumption about djam.dylan. I'll amend it with djam.dylan changed and see if it bootstraps.

PR #1569, opened 2024-03-05
Remove unused submodules

cgay:
There's no need to bundle libraries into OD as submodules unless they're needed by OD itself.


cgay:
Submodules.  Sigh.

PR #1585, opened 2024-05-02
Unified executables fixes

housel:
These changes ensure that runtime libraries are copied to the target for unified executables, and ensure that symbols in unified executables are still exported to allow backtraces.


cgay:
Add a release note?

waywardmonkeys (commented):
Are these the same across Linux, FreeBSD, and macOS toolchains?

cgay (approved):

PR #1613, opened 2024-06-10
Begin deprecating <format-string-condition> in favor of <simple-condition>

cgay:
Begin deprecating `<format-string-condition>` in favor of `<simple-condition>`. There's no point in having two names for the same class.


cgay:
I'll redo this after submitting a PR to fix `<type-error>`.

PR #1617, opened 2024-06-14
doc: Add Platforms: keyword to LID File Format document

cgay:


fraya:
netbsd platform is working?

cgay:
> netbsd platform is working?

I'm not sure if it's working these days (and we haven't been building the releases for it) but it's one of the supported platform identifiers.

PR #1618, opened 2024-06-19
If-merge optimizations

housel:
These changes improve the compiler's type estimation of certain types of if expressions.


cgay:
Is it possible to make a small enough example of what the improvement does to fit in the release notes?

cgay (approved):

PR #1621, opened 2024-06-23
Dispatch profiling

housel:
These changes fix dispatch profiling, make it more usable in practice, and document how to use it.

### DIFF
--- a/documentation/source/library-reference/dispatch-profiler/index.rst
+++ b/documentation/source/library-reference/dispatch-profiler/index.rst
@@ -1,0 +1,261 @@
+*****************************
+The DISPATCH-PROFILER library
+*****************************
+
+.. current-library:: dispatch-profiler
+
+The dispatch profiler library exposes the multimethod dispatch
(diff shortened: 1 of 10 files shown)


documentation/source/library-reference/dispatch-profiler/index.rst
```diff
@@ -0,0 +1,261 @@
...
+  $ OPEN_DYLAN_PROFILE_ALL_CALLS=1 dylan-compiler -build write-100mb
+
+
+Report Output Format
+********************
+
+After a profiling run, upon collection of dispatch statistics a
+profiling report can be written out. The report, sorted by ``COST``
+field, lists generic functions and call sites for which and method
+invocations present in the dispatch cache. The following fields appear
```
cgay, line 41: Should this be "for which ANY method invocations ARE present" ?

cgay (changes requested):

cgay (approved):

PR #1626, opened 2024-09-21
common-dylan/timers.dylan: new function microsecond-counter

cgay:
I needed some kind of low-level system counter to implement a clock() built-in while writing the interpreter in Crafting Interpreters and common-dylan:timers has almost the right thing and seems a natural place to add one.

This is for review...I'll modify the comments at a minimum before merging but for now they reflect some of my thought process.

### DIFF
--- a/sources/common-dylan/library.dylan
+++ b/sources/common-dylan/library.dylan
@@ -47,7 +47,8 @@ define module simple-timers
          timer-start,
          timer-stop,
          timer-accumulated-time,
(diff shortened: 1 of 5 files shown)


cgay:
This doesn't feel right yet.  I think a timer _should_ use nanoseconds. I could truncate the seconds to some arbitrary time in the past that _would_ fit in a single integer, maybe? But it's not that important so I'm just closing this.

PR #1636, opened 2024-12-12
Multi assignment DEP

cgay:


documentation/source/proposals/dep-0013-multi-assignment.rst
```diff
@@ -0,0 +1,236 @@
...
+       let (seconds, microseconds) = timer-stop(state[#"cpu-profiling-timer"]);
+       state[#"cpu-time-seconds"]      := seconds;
+       state[#"cpu-time-microseconds"] := microseconds;
+       state[#"cpu-profiling"]         := #f;
+       state[#"cpu-profiling-timer"]   := #f;
+     end
+   end method stop-profiling-type;
+
+   // new:
+   define method stop-profiling-type
```
fraya, line 128: Maybe this can be an example of more places than values:

```
define method stop-profiling-type
       (state :: <profiling-state>, keyword :: <cpu-profiling-type>) => ()
     when (element(state, #"cpu-profiling", default: #f))
      (state[#"cpu-time-seconds"],
       state[#"cpu-time-microseconds"],
       state[#"cpu-profiling"],
       state[#"cpu-profiling-timer"])
         := timer-stop(state[#"cpu-profiling-timer"]);
     end
 end method stop-profiling-type;
```

cgay: I would personally avoid doing that because if `timer-stop` is changed such that it returns *more* values (a change that is normally safe to make), this code is likely to break.

PR #1662, opened 2025-02-18
Lexer love

cgay:
This is a fairly big change to the lexer. See commit logs for details. Overview:
* Add a new kind of state that allows side-effecting the lexer to keep track of state.
* Add `#"otherwise"` state transition spec to cover any characters not explicitly listed.
* Add a `#"not-allowed"` state. Explicit transfer to this state in conjunction with `#"otherwise"` can be much clearer than trying to figure out which characters have no transition.
* Fix bugs in multi-line string literal parsing and allow delimiters of any length >= 3.
* Lex comments (both single-line and delimited) using the state machine instead of the special `skip-multi-line-comment` function.
* The inevitable cleanups


sources/dfmc/reader/tests/literal-test-suite.dylan
```diff
@@ -19,13 +19,15 @@ define function verify-presentation
...
+/*
   verify-literal(read-fragment("#b10"), 2, <integer-fragment>);
   verify-literal(read-fragment("#b010"), 2, <integer-fragment>);
   verify-literal(read-fragment("#b1111_0000"), 240, <integer-fragment>);
+*/
   assert-false(read-fragment("#b10_"));
-  assert-signals(<invalid-token>, read-fragment("#b_10"));
-  assert-signals(<invalid-token>, read-fragment("#b2"));
-  assert-signals(<invalid-token>, read-fragment("#b + 1"));
+//  assert-signals(<invalid-token>, read-fragment("#b_10"));
```
housel, line 28: Why are these commented out?

housel (approved):

sources/dfmc/reader/tests/literal-test-suite.dylan
cgay: Fixed by uncommenting them. Looks like something I broke at one point and was debugging by commenting out all but one assertion. Thanks.

PR #1667, opened 2025-03-07
Remove collections convenience module, doc updates

cgay:


documentation/source/library-reference/common-dylan/byte-vector.rst
```diff
@@ -14,6 +15,8 @@ The byte-vector Module
 
      - :type:`<byte>`
 
+.. <byte> is defined in the dylan library but documented here (for reasons?).
```
housel, line 18: It's defined in the dylan library because it's referenced by language extensions and implementation details that have to be there.

housel (approved):

documentation/source/library-reference/common-dylan/byte-vector.rst
cgay: Right, I was a bit lazy with that comment and I'll remove it.

I was basically wondering why we don't document it in `dylan-extensions` directly instead of somewhere that it is re-exported from.

PR #1669, opened 2025-03-08
doc: Add a note about add! and remove! differences from DRM

cgay:


housel (approved):

cgay:
The force-push for this PR was to format the two generic functions the same way and to fix grammar.

PR #1671, opened 2025-03-09
dfmc-reader: fix handling of names with leading numeric

### LINKED ISSUE #1395: Lexical grammar does not follow DRM
According to the DRM a [name](https://opendylan.org/books/drm/Lexical_Grammar#IX-2101) is allowed to start with a number (subject to other conditions) but this is not always the case. For example,
```dylan
define constant 0xx = 23;
```
is allowed but
```dylan
define constant 0yy = 24;
```
gives 
```
Serious warning - Invalid syntax in constant-definer macro call.

/tmp/parser/parser.dylan:79.1-25: Serious warning - Skipping constant-definer macro call due to previous syntax error.
      ------------------------
      define constant 0yy = 24;
      ------------------------
```
This seems to depend on the first letter in the proposed name, specifically 'd', 'e', 's', 'x' are allowed, others give an error. (for example `123elephant` would be allowed as the first letter is 'e')

Note that names starting with numbers are not recommended and rarely if ever used in Dylan code.

cgay:
Fixes #1395


sources/dfmc/reader/lexer-transitions.dylan
```diff
@@ -573,7 +573,8 @@ define constant $initial-state :: <state>
          #('/' . #"decimal-slash"),
          #('.' . #"decimal-dot"),
          #("eEsSdDxX" . #"decimal-e"),
-         #("_" . #"decimal-underscore")),
+         #("_" . #"decimal-underscore"),
+         #("ABCF-RT-WYZabcf-rt-wyz" . #"numeric-alpha")),
```
housel, line 577: Could we move this line up one so it's adjacent to the other letters?

housel (approved):

sources/dfmc/reader/lexer-transitions.dylan
cgay: Done

PR #1686, opened 2025-04-07
doc: Add cheetsheet for python

### LINKED ISSUE #1665: Python to Dylan cheat sheet
Would be nice to have a Python to Dylan cheat sheet.

fraya:
Closes #1665


documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|    x == y                      |    x == y                                     |
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | N/A                                           |
+|                                |                                               |
+|    x is y                      |                                               |
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    x == y                      |    x = y                                      |
++--------------------------------+-----------------------------------------------+
```
cgay, line 229: Python `is` is closer to Dylan `==` and Python `==` is closer to Dylan `=`.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+
+Pairs and Lists
+^^^^^^^^^^^^^^^
```
cgay, line 232: TBH a lot of this section doesn't feel very helpful for Pythonistas. They won't really be looking for a linked list implementation and some of the equivalences feel a bit contrived.  It would work better in a "Python for Dylaners" document.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|                                |                                               |
+|    obj in ls                   |    member?(obj, ls)                           |
++--------------------------------+-----------------------------------------------+
+| N/A                            | .. code-block:: dylan                         |
+|                                |                                               |
+|                                |    member?(obj, ls, test: \==)                |
++--------------------------------+-----------------------------------------------+
+
+Symbols
+^^^^^^^
```
cgay, line 310: Similar here.  Python folks won't be looking for a way to use symbols and `str` isn't really the same thing.  Possibly if you switch to using `sys.intern` instead it will work.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|        obj == int(obj))          |                                               |
++----------------------------------+-----------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                         |
+|                                  |                                               |
+|    isinstance(obj, int)          |    instance?(obj, <integer>)                  |
+|                                  |    integral?(num)                             |
++----------------------------------+-----------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                         |
+|                                  |                                               |
+|    n1 = n2                       |    n1 = n2                                    |
```
cgay, line 359: You want `==` on the Python side.  And probably should delete `=` on the Dylan side since in this context we're comparing numbers only.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|                                  |                                               |
+|    n == 0                        |    zero?(n)                                   |
++----------------------------------+-----------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                         |
+|                                  |                                               |
+|    n > 0                         |    positive?(n)                               |
++----------------------------------+-----------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                         |
+|                                  |                                               |
+|    n < 0                         |    negative?(n)                               |
```
cgay, line 388: The above three should show that `==`, `>`, and `<` also work in Dylan. You might want to somehow show that `0 == 0.0` is false in Dylan.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|                                  |                                               |
+|    1 * 2 * 3                     |    1 * 2 * 3                                  |
++----------------------------------+-----------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                         |
+|                                  |                                               |
+|    5 - 3                         |    5 - 3                                      |
++----------------------------------+-----------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                         |
+|                                  |                                               |
+|    2.3 / 1.7                     |    2.3 / 1.7                                  |
```
cgay, line 412: An important difference is that `1 / 2` is an error in Dylan, so that would be good to show.  

```
1 / 2 => 0.5         1 / 2 => error
                     1.0 / 2 => 0.5
                     truncate/(1, 2) => 0
                     ceiling/(1, 2) => 1
                     floor/(1, 2) => 0
```

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
++================================+===============================================+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    isinstance(obj, str) and    |    instance?(obj, <character>)                |
+|      len(obj) == 1             |                                               |
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    char1 == char2              |    char1 = char2                              |
+|                                |    char1 == char2                             |
```
cgay, line 446: Remove the `=` case.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|    isinstance(obj, str)        |    instance?(obj, <string>)                   |
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    char * k                    |    make(<string>, size: k, fill: char)        |
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | N/A                                           |
+|                                |                                               |
+|    "".join([char for char in   |                                               |
+|         [char, ...]])          |                                               |
```
cgay, line 497: What's the purpose of this example?  To create a string from a sequence of characters in Dylan is just `as(<string>, #['a', 'b', 'c'])`  Maybe the Python side should just be `"".join(['a', 'b', 'c'])`

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
++--------------------------------+-----------------------------------------------+
+| Python                         | Dylan                                         |
++================================+===============================================+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    isinstance(obj, str)        |    instance?(obj, <string>)                   |
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    char * k                    |    make(<string>, size: k, fill: char)        |
```
cgay, line 492: Maybe replace both `char` with `'x'` since Python doesn't have characters.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|                                |                                               |
+|    isinstance(obj, list)       |    instance?(obj, <vector>)                   |
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    [fill] * k                  |    make(<vector>, size: k, fill: fill)        |
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    [obj ...]                   |    vector(obj, ...);                          |
```
cgay, line 563: The Python side needs a comma.  The Dylan side doesn't really need the semicolon.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    # Uses lists for both       |    as(<vector>, list)                         |
+|    list(list)                  |                                               |
++--------------------------------+-----------------------------------------------+
+| .. code-block:: python         | .. code-block:: dylan                         |
+|                                |                                               |
+|    i in range(len(vec)):       |    fill!(vec, obj)                            |
+|      vec[i] = obj              |                                               |
```
cgay, line 592: I think the Python side is missing a `for`.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|        pass                       |    end block                                  |
+|      body(k)                      |                                               |
++-----------------------------------+-----------------------------------------------+
+| .. code-block:: python            | .. code-block:: dylan                         |
+|                                   |                                               |
+|    def example():                 |    block (k)                                  |
+|      try:                         |      body                                     |
+|        body()                     |    cleanup                                    |
+|      finally:                     |      cleanup-stuff                            |
+|        cleanup_stuff()            |    end block                                  |
```
cgay, line 653: Add an `except` / `exception` clause to each?

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+
++-----------------------------------+-----------------------------------------------+
+| Python                            | Dylan                                         |
++===================================+===============================================+
+| .. code-block:: python            | .. code-block:: dylan                         |
+|                                   |                                               |
+|    def example():                 |    block (k)                                  |
+|      def k():                     |      body                                     |
+|        pass                       |    end block                                  |
+|      body(k)                      |                                               |
```
cgay, line 645: Not sure what you want to express here.  ?

cgay (changes requested):
This is great stuff.  I hope you're not put off by all the comments; for such a large change it's probably inevitable. :)

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|    'Hello\n'           |    "Hello\n"          |
++------------------------+-----------------------+
+| .. code-block:: python | .. code-block:: dylan |
+|                        |                       |
+|    'apple'             |    #"apple"           |
+|                        |    apple:             |
++------------------------+-----------------------+
+| .. code-block:: python | .. code-block:: dylan |
+|                        |                       |
+|    'two words'         |    #"two words"       |
```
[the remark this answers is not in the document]
fraya: I have read that Python interns strings, 

```
> a = 'foo'
> b = 'foo'
> a is foo
> True
```
https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is
What do you think?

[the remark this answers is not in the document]
fraya: I agree that 'Foo' is not equal to 'foo'.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+|                        |                       |
+|    '\n'                |    '\n'               |
++------------------------+-----------------------+
+| .. code-block:: python | .. code-block:: dylan |
+|                        |                       |
+|    'Hello'             |    "Hello"            |
++------------------------+-----------------------+
+| .. code-block:: python | .. code-block:: dylan |
+|                        |                       |
+|    'Hello\n'           |    "Hello\n"          |
```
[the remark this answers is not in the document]
fraya: Python use triple single quotes (https://www.geeksforgeeks.org/multiline-string-in-python/).
```
>>> say = '''Hey girls, B-boys
... Superstar DJs, here we go!
... '''
>>> print(say)
Hey girls, B-boys
Superstar DJs, here we go!

>>> say = '''Hey girls, B-boys
...          Superstar DJS, here we go!'''
>>> print(say)
Hey girls, B-boys
         Superstar DJS, here we go!
```
This means that Python don't follow the "Rectangle rule"?

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
+   In Dylan, any words after an ``end`` (e.g.  ``end method``) are
+   optional.
+
++----------------------------------+---------------------------------------+
+| Python                           | Dylan                                 |
++==================================+=======================================+
+| .. code-block:: python           | .. code-block:: dylan                 |
+|                                  |                                       |
+|    var = exp                     |    define variable var = exp          |
+|    VAR = exp # Convention        |    define constant var = exp          |
```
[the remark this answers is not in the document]
fraya: Why not?
I think we can put two lines, one without type and another with type.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,654 @@
...
++----------------------------------+---------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                 |
+|                                  |                                       |
+|    f(x, y, z)                    |    f(x, y, z)                         |
```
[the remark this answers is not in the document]
fraya: Can you provide an example?

documentation/source/documentation/cheatsheets/python.rst
fraya: My fault in left side, but I don't understand why the right side is wrong.

documentation/source/documentation/cheatsheets/python.rst
fraya: Is "Lost in Python translation". What is the equivalent of the `block` in Opendylan in Python?

documentation/source/documentation/cheatsheets/python.rst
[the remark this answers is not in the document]
fraya: Another example. Python puts a newline at the beginning
```
>>> s = '''
...     Hello, from
...     Python
...     '''
>>> print(s)

    Hello, from
    Python
```

documentation/source/documentation/cheatsheets/python.rst
fraya: `sys.intern` 
```
>>> a = sys.intern('a')
>>> b = sys.intern('A')
>>> a is b
False
``

File: documentation/source/documentation/cheatsheets/python.rst [hunk 18]
```diff
@@ -0,0 +1,654 @@
...
+|                        |                       |
+|    0o644               |    #o644              |
++------------------------+-----------------------+
+| .. code-block:: python | .. code-block:: dylan |
+|                        |                       |
+|    0x2A5F              |    #x2A5F             |
++------------------------+-----------------------+
+| .. code-block:: python | .. code-block:: dylan |
+|                        |                       |
+|    6.02e+23            |    6.02E23            |
```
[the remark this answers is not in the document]
cgay: I think just leave out the `e` syntax for Dylan since it could (theoretically) be repurposed for extended precision floats.

documentation/source/documentation/cheatsheets/python.rst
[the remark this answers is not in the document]
cgay: It interns string _literals_ but not strings in general, so unless you're really careful it's not a good idea to use `is` for strings.

documentation/source/documentation/cheatsheets/python.rst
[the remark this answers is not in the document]
cgay: Yeah, looks like Python doesn't follow the Rectangle Rule.

It accepts both `'''` and `"""` as delimiters.

documentation/source/documentation/cheatsheets/python.rst
cgay: There's nothing directly equivalent to `block`, but Pythonistas will be interested in how to do `try/except/finally`, which has direct equivalents in `block`.

You could add [a non-local exit example](https://play.opendylan.org/shared/a32e7c1a359f8753) and put `N/A` on the Python side.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,600 @@
...
+| .. code-block:: python           | .. code-block:: dylan                    |
+|                                  |                                          |
+|    1                             |    begin 1; 2; 3; end                    |
+|    2                             |    begin 1; 2; 3 end                     |
+|    3                             |                                          |
++----------------------------------+------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                    |
+|                                  |                                          |
+|    lambda x, y, *z:              |    method (x, y, #rest z)                |
+|      print('hello')              |      say("hello");                       |
```
cgay, line 117: `say` should be `format-out`.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,600 @@
...
+|      return f(x, y, z)           |      f(x, y, z)                          |
+|                                  |    end method                            |
++----------------------------------+------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                    |
+|                                  |                                          |
+|    x = 5                         |    begin                                 |
+|    body                          |      let x = 5;                          |
+|                                  |      body                                |
+|                                  |    end                                   |
```
cgay, line 126: This is a bit unfair to Dylan. If your intention is to show how to create a local scope, then on the Python side I would just say "The function is the smallest scope".  The only way to do it (unless there's something new in Python 3?) is to define a local function and call it.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,600 @@
...
+|      then1                       |      then1;                              |
+|      then2                       |      then2;                              |
+|    else:                         |    else                                  |
+|      else1                       |      else1;                              |
+|      else2                       |      else2;                              |
+|                                  |    end if                                |
++----------------------------------+------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                    |
+|                                  |                                          |
+|    var = value                   |    var := value                          |
```
cgay, line 161: How about removing this row and including both `var := value` and `let var = value` in the first row of this table with `define constant` et al?

Python really overloads the hell out of the `=` operator.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,600 @@
...
+|    elif test2:                   |      test2 => result2;                   |
+|      result2                     |      otherwise => result                 |
+|    else:                         |    end case                              |
+|      result                      |                                          |
++----------------------------------+------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                    |
+|                                  |                                          |
+|    if exp in ('a', 2):           |    select (exp)                          |
+|      result1                     |      #"a", 2 => result1;                 |
+|    elif exp in ('a' 'b'):        |      'a', 'b' => result2;                |
```
cgay, line 184: The Python side needs a comma.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,600 @@
...
+|      result2                     |      otherwise => result                 |
+|    else:                         |    end select                            |
+|      result                      |                                          |
++----------------------------------+------------------------------------------+
+| .. code-block:: python           | .. code-block:: dylan                    |
+|                                  |                                          |
+|    c = comparison-func(exp)      |    select (exp by comparison-func)       |
```
cgay, line 191: `comparison-func` isn't really doing comparison here. How about calling it `f`?

documentation/source/documentation/cheatsheets/python.rst
cgay: That's what I would expect.

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,600 @@
...
+
++-----------------------------------+-----------------------------------------------+
+| Python                            | Dylan                                         |
++===================================+===============================================+
+| .. code-block:: python            | .. code-block:: dylan                         |
+|                                   |                                               |
+|    def example():                 |    block (k)                                  |
+|      try:                         |      body                                     |
+|        body()                     |    exception (e :: <error>)                   |
+|      except Error:                |      handle-error()                           |
```
cgay, line 580: `except Error as e:` is closer to the Dylan side.  (Or you can change the Dylan side to `exception (<error>)`.)

documentation/source/documentation/cheatsheets/python.rst
```diff
@@ -0,0 +1,600 @@
...
+|        cleanup_stuff()            |    end block                                  |
++-----------------------------------+-----------------------------------------------+
+| N/A                               | .. code-block:: dylan                         |
+|                                   |                                               |
+|                                   |    define function top ()                     |
+|                                   |      block (exit-block)                       |
+|                                   |        bar(exit-block);                       |
+|                                   |        format-out("You won't see this.\n");   |
+|                                   |      end;                                     |
+|                                   |      format-out("Top done\n");                |
```
cgay, line 592: In this context I guess it would be clearer to say `format-out("You WILL see this.\n")`

PR #1689, opened 2025-04-08
Include detailed Git info in Open Dylan version string

cgay:
This uses the same mechanism as Deft to maintain version info, so `dylan-compiler -version` displays something similar to "v2025.1.0-13-g7f7fdb9 built on 2025-01-08T21:05:44-05:00".

Besides more detail in the version, making it easier to pinpoint the commit from which Open Dylan was built, this also makes it so that we don't have to manually update the version string.

The release tag must be applied before the build, which is a minor change in how we've traditionally done releases.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# To actually use this filter you must run the following commands in the top-level repo
+# directory:
+#   git config filter.version.smudge ./tools/scripts/version-smudge.sh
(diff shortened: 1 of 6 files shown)


cgay:
```
$ dylan-compiler -version
v2024.1.0-881-g49b49165c built on 2025-04-08T14:09:30-04:00
```

PR #1713, opened 2025-05-11
feat: Add 'symbolic-link' method for UNIX

fraya:
Adds UNIX method to create a symbolic link in the module 'file-system' of 'system' library.

### DIFF
--- a/documentation/source/library-reference/system/file-system.rst
+++ b/documentation/source/library-reference/system/file-system.rst
@@ -33,6 +33,7 @@ set any available properties for the file.
 - :func:`file-property-setter`
 - :gf:`expand-pathname`
 - :gf:`shorten-pathname`
+- :gf:`create-symbolic-link`
 - :macro:`with-open-file`
 
 Manipulating directories
@@ -407,6 +408,49 @@ File-System module.
 
      - :func:`delete-directory`
 
+.. generic-function:: create-symbolic-link
+
+   Creates a symbolic link
+
+   :signature: create-symbolic-link *target* *link* => ()
+
+   :param target: An instance of `<pathname>`:class:.
+   :param link:   An instance of `<pathname>`:class:.
+
+   :description:
+
+     Creates a symbolic link to *target*.
+
+   :example:
+
+     Creates a symbolic link named *whattimeislove* to the *date* 
+     utility.
+
+     .. code-block:: dylan
+
+        create-symbolic-link("/usr/bin/date", "/tmp/whattimeislove"); 
+
+.. method:: create-symbolic-link
+   :specializer: <pathname>, <pathname>
+   :no-contents-entry:
+
+   :parameter target: An instance of :class:`<pathname>`.
+   :parameter link:   An instance of :class:`<pathname>`.
+
+.. method:: create-symbolic-link
+   :specializer: <string>, <string>
+   :no-contents-entry:
+
+   :parameter target: An instance of :drm:`<string>`.
+   :parameter link:   An instance of :drm:`<string>`.
+
+.. method:: create-symbolic-link
+   :specializer: <file-system-locator>, <file-system-locator>
+   :no-contents-entry:
+
+   :parameter target: An instance of :class:`<file-system-locator>`.
+   :parameter link:   An instance of :class:`<file-system-locator>`.
+
 .. function:: delete-directory
 
    Deletes the specified directory.
@@ -458,7 +502,7 @@ File-System module.
       and ".." directories are not included in the result.
 
 .. generic-function:: directory-empty?
-   
+
    Checks whether a directory is empty or not.
 
    :signature: directory-empty? *directory* => *empty?*
@@ -471,7 +515,7 @@ File-System module.
 
    :param directory: An instance of :class:`<file-system-directory>`.
    :value empty?: An instance of :class:`<boolean>`.
-   
+
 .. function:: do-directory
 
    Executes the supplied function once for each entity in the specified
@@ -917,6 +961,10 @@ File-System module.
       Repeatedly follows symbolic links starting with *file* until it finds a
       non-link file or directory, or a non-existent link target.
 
+   :seealso:
+
+     - :gf:`create-symbolic-link`
+
 .. _make:
 
 .. method:: make
@@ -1074,6 +1122,7 @@ File-System module.
      - :func:`home-directory`
      - :func:`link-target`
      - :func:`rename-file`
+     - :gf:`create-symbolic-link`
 
 .. class:: <posix-file-system-locator>
    :abstract:
--- a/documentation/source/library-reference/system/file-system.rst
+++ b/documentation/source/library-reference/system/file-system.rst
@@ -420,6 +420,7 @@ File-System module.
    :description:
 
      Creates a symbolic link to *target*.
+     Not supported in Win32.
 
    :example:
 
--- a/documentation/source/library-reference/system/file-system.rst
+++ b/documentation/source/library-reference/system/file-system.rst
@@ -408,14 +408,14 @@ File-System module.
 
      - :func:`delete-directory`
 
-.. generic-function:: create-symbolic-link
+.. function:: create-symbolic-link
 
    Creates a symbolic link
 
    :signature: create-symbolic-link *target* *link* => ()
 
-   :param target: An instance of `<pathname>`:class:.
-   :param link:   An instance of `<pathname>`:class:.
+   :param target: An instance of :class:`<pathname>`.
+   :param link:   An instance of :class:`<pathname>`.
 
    :description:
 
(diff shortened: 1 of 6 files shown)


sources/system/file-system/file-system.dylan
```diff
@@ -516,3 +516,19 @@ define sideways method list-locator
      locator);
   locators
 end method list-locator;
+
+///
+
+define generic symbolic-link
```
cgay, line 522: This could use a test.  See also `test-file-exists?`.  You can use `test-temp-directory`.  Looks like I was lazy and didn't write a test using `run-application("ln -s ...)`.

documentation/source/library-reference/system/file-system.rst
```diff
@@ -1179,6 +1185,42 @@ File-System module.
...
+
+     Creates a symbolic link named *whattimeislove* to the *date* 
+     utility.
+
+     .. code-block:: dylan
+
+        symbolic-link("/usr/bin/date", "/tmp/whattimeislove"); 
+
+.. method:: symbolic-link
+   :specializer: <string>
```
cgay, line 1211: I think this needs to be `:specializer: <string>, <string>`. Also below.

sources/system/file-system/file-system.dylan
```diff
@@ -516,3 +516,19 @@ define sideways method list-locator
...
+  (target :: <pathname>, link :: <pathname>) => ();
+
+define method symbolic-link
+    (target :: <file-system-locator>, link :: <file-system-locator>) => ()
+  %symbolic-link(target, link)
+end method symbolic-link;
+
+define method symbolic-link 
+    (target :: <string>, link :: <string>) => ()
```
cgay, line 531: Specialize on `<pathname>, <pathname>` here to cover all the dispatch cases.

cgay (changes requested):

documentation/source/library-reference/system/file-system.rst
```diff
@@ -33,6 +33,7 @@ set any available properties for the file.
 - :func:`file-property-setter`
 - :gf:`expand-pathname`
 - :gf:`shorten-pathname`
+- :gf:`symbolic-link`
```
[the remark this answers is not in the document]
fraya: Ok, I'll rename it. Should I rename `%symbolic-link` too?

If you don't mind I'll make `create-hard-link` in another PR.

sources/system/file-system/file-system.dylan
fraya: I don't understand it. `<pathname>` is either a `<string>` or a `<file-system-locator>`. I have specialize the gf in both types.

documentation/source/library-reference/system/file-system.rst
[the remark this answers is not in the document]
cgay: > Ok, I'll rename it. Should I rename `%symbolic-link` too?

Probably, yes.

> If you don't mind I'll make `create-hard-link` in another PR.

Absolutely.

sources/system/file-system/file-system.dylan
cgay: You are covering `string X string` and `locator X locator` but there is also `string X locator` and `locator X string`.  

On second thought you can actually do it all with one method on `pathname X pathname` as long as you call `as(<file-system-locator>, arg)` on each method argument.

documentation/source/library-reference/system/file-system.rst
```diff
@@ -407,6 +408,49 @@ File-System module.
...
      - :func:`delete-directory`
 
+.. generic-function:: create-symbolic-link
+
+   Creates a symbolic link
+
+   :signature: create-symbolic-link *target* *link* => ()
+
+   :param target: An instance of `<pathname>`:class:.
+   :param link:   An instance of `<pathname>`:class:.
```
cgay, line 418: Does this syntax work?  We usually write ```:class:`<pathname>` ```

documentation/source/library-reference/system/file-system.rst
```diff
@@ -407,6 +408,49 @@ File-System module.
...
+     Creates a symbolic link to *target*.
+
+   :example:
+
+     Creates a symbolic link named *whattimeislove* to the *date* 
+     utility.
+
+     .. code-block:: dylan
+
+        create-symbolic-link("/usr/bin/date", "/tmp/whattimeislove");
```
cgay, line 431: Nit: no need for a trailing semicolon

sources/system/file-system/file-system.dylan
```diff
@@ -516,3 +516,25 @@ define sideways method list-locator
...
+define method create-symbolic-link
+    (target :: <file-system-locator>, link :: <file-system-locator>) => ()
+  %create-symbolic-link(target, link)
+end method create-symbolic-link;
+
+define method create-symbolic-link
+    (target :: <string>, link :: <string>) => ()
+  create-symbolic-link(as(<file-system-locator>, target),
+                       as(<file-system-locator>, link))
+end method create-symbolic-link;
```
cgay, line 534: Delete the above two methods, which are redundant with the method below.

Optional: since this is not an open generic, and the "dispatch" to OS-specific code is done via LID files, you could just define it with `define function`.

documentation/source/library-reference/system/file-system.rst
```diff
@@ -407,6 +408,49 @@ File-System module.
...
+
+   :parameter target: An instance of :drm:`<string>`.
+   :parameter link:   An instance of :drm:`<string>`.
+
+.. method:: create-symbolic-link
+   :specializer: <file-system-locator>, <file-system-locator>
+   :no-contents-entry:
+
+   :parameter target: An instance of :class:`<file-system-locator>`.
+   :parameter link:   An instance of :class:`<file-system-locator>`.
```
cgay, line 452: Doc needs updating depending on how you resolve my comments below about these methods.

documentation/source/library-reference/system/file-system.rst
```diff
@@ -1074,6 +1122,7 @@ File-System module.
      - :func:`home-directory`
      - :func:`link-target`
      - :func:`rename-file`
+     - :gf:`create-symbolic-link`
```
cgay, line 1125: This will need to be `:func:` if you take my suggestion below.

sources/system/tests/file-system.dylan
```diff
@@ -69,14 +69,16 @@ define test test-file-exists? ()
...
+  let tmp-nothing  = file-locator(dir, "nothing");
+  let symlink-bad  = file-locator(dir, "symlink-bad");
+  let symlink-good = file-locator(dir, "symlink-good");
+
+  create-symbolic-link(tmp-nothing, symlink-bad);
+  create-symbolic-link("/usr", symlink-good);
+
+  assert-true(file-exists?(symlink-good));
+  assert-false(file-exists?(symlink-bad));
+  assert-true(file-exists?(symlink-bad, follow-links?: #f));
```
cgay, line 81: Thank you for fixing my mess.  :)

documentation/source/library-reference/system/file-system.rst
fraya: Yes, hmmm. The Yoda way of documentation, it is. Correct it, I will.

documentation/source/library-reference/system/file-system.rst
fraya: Done

sources/system/file-system/file-system.dylan
fraya: You only want this function?
```
define function create-symbolic-link
    (target :: <pathname>, link :: <pathname>) => ()
  %create-symbolic-link(as(<file-system-locator>, target),
                        as(<file-system-locator>, link))
end function create-symbolic-link;
```
If the answer is yes, perhaps there is more code in this file that can be reviewed the same way.

cgay: Right. Unless I'm very confused the other methods are totally redundant. And yes, I see other generic functions that could be turned into regular functions, but in a different PR.

One could argue that using specific methods is more correct because there's no need to call `as(<file-system-locator>, ...)` on arguments that are already locators, but I don't see how it's worth it.  Maybe @housel has a different opinion?

fraya: Then I think the code is ready to merge.

cgay (approved):

sources/system/file-system/file-system.dylan
housel: This is going to break the build on Win32. Even if we don't add support for actually creating symbolic links on Windows (which uses the `CreateSymbolicLink` API, which is not included in the link libraries we're using to build Open Dylan), we should at least leave a placeholder for `%create-symbolic-link` so that the compile will not fail.

fraya: @housel Thank you. I added the placeholder in my last commit. Can you review it?

sources/system/file-system/win32-file-system.dylan
```diff
@@ -626,3 +626,9 @@ define function %root-directories
     end
   end
 end function %root-directories;
+
+define function %create-symbolic-link
+    (target :: <microsoft-directory-locator>, link :: <microsoft-directory-locator>)
+ => ()
+  win32-file-system-error("symlink", "create symbolic link. Not supported in Win32")
```
housel, line 633: It should be enough to just call `error` here; the `win32-file-system-error` function retrieves the Win32 equivalent of `errno` and includes that in the `error` message.

housel (approved):

PR #1736, opened 2025-07-07
feat: Update '%copy-file' shell copy

### LINKED ISSUE #1649: copy-file is implemented by shelling out to /bin/cp on Unix platforms
Modern Linux and FreeBSD have a `copy_file_range` system call, and mac OS has `copyfile`; these can be used directly to implement `%copy-file` rather than using `run-application` to invoke the `cp` program.

fraya:
Replace shell copy with specialized copy.

Use 'copy_file_range' in Unix with fallback copy 'unix-copy-file' in case of error (e.g. cross-device copy).

Closes #1649

### DIFF
--- a/sources/system/file-system/unix-file-system.dylan
+++ b/sources/system/file-system/unix-file-system.dylan
@@ -203,48 +203,48 @@ define function %copy-file
           let file-descriptor
             = unix-open(as(<string>, locator), flags-mode, flags-create);
           if (file-descriptor < 0)
-            let errno = unix-errno();
-            if (errno = $e_access)
-              error(make(<invalid-file-permissions-error>,
-                         locator: as(<posix-file-locator>, locator)))
-            else
-              unix-file-error("open", "file %s", as(<string>, locator))
-            end
-          end;
+            select (unix-errno())
+              $e_access =>
+                error(make(<invalid-file-permissions-error>,
+                           locator: as(<posix-file-locator>, locator)));
+              $e_exists =>
+                error(make(<file-exists-error>,
+                           locator: as(<posix-file-locator>, locator)));
+              otherwise =>
+                unix-file-error("open", "file %s", as(<string>, locator));
+            end select;
+          end if;
           file-descriptor
-        end,
+        end method,
         method open-source-file(filename)
-          let flags-mode   = $O_RDONLY;
-          let flags-create = $O_RDONLY;
+          let flags-mode   = $o_rdonly;
+          let flags-create = 0;
           open-file(filename, flags-mode, flags-create);
-        end,
-        method open-destination-file(filename)
-          let flags-mode   = logior($O_WRONLY, $O_CREAT, $O_TRUNC);
-          let flags-create = logior($S_IRUSR, $S_IWUSR, $S_IWGRP, $S_IROTH);
+        end method,
+        method open-destination-file(filename, if-exists)
+          let flags-mode 
+            = if (if-exists = #"signal")
+                logior($o_wronly, $o_creat, $o_trunc, $o_excl)
+              else 
+                logior($o_wronly, $o_creat, $o_trunc)
+              end;
+          let flags-create 
+            = logior($s_irusr, $s_iwusr, $s_irgrp, $s_iwgrp, $s_iroth, $s_iwoth);
           open-file(filename, flags-mode, flags-create);
         end method;
 
   let source = %expand-pathname(source);
   let destination = %expand-pathname(destination);
-  // UNIX strikes again!  The copy command will overwrite its target if
-  // the user has write access and the only way to prevent it would
-  // require the user to respond to a question!  So, we have to manually
-  // check beforehand.  (Just another reason I'm a member of Unix-Haters)
-  if (if-exists = #"signal" & file-exists?(destination))
-    error(make(<file-system-error>,
-               format-string: "File exists: Can't copy %s to %s",
-               format-arguments: list(as(<string>, source),
-                                      as(<string>, destination))))
-  end;
 
-  let source-size    :: <integer> = file-property(source, #"size");
   let source-fd      :: <integer> = -1;
   let destination-fd :: <integer> = -1;
 
   block ()
 
     source-fd      := open-source-file(source);
-    destination-fd := open-destination-file(destination);
+    destination-fd := open-destination-file(destination, if-exists);
+
+    let source-size :: <integer> = file-property(source, #"size");
 
     let failed?
       = primitive-raw-as-boolean
@@ -260,15 +260,24 @@ define function %copy-file
 
     // fallback copy in case of error
     if (failed?)
-      unix-copy-file(source-fd, destination-fd);
+      let status = unix-copy-file(source-fd, destination-fd);
+      unless (status = #"ok")
+        unix-file-error(if (status = #"read-error") "read" else "write" end,
+                        "copy from file %s to %s",
+                        source, destination)
+      end;
     end;
 
   cleanup
     unless (source-fd < 0)
-      unix-close(source-fd)
+      if (unix-close(source-fd) = -1)
+        unix-file-error("close", "close file %s", source)
+      end
     end;
     unless (destination-fd < 0)
-      unix-close(destination-fd)
+      if (unix-close(destination-fd) = -1)
+        unix-file-error("close", "close file %s", destination)
+      end
     end;
   end block;
 end function %copy-file;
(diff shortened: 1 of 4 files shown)


sources/system/file-system/unix-interface.dylan
```diff
@@ -106,6 +123,32 @@ define function unix-delete-file (path :: <byte-string>) => (ok :: <boolean>)
...
+define function unix-copy-file
+    (source-fd :: <integer>, destination-fd :: <integer>, #key block-size = 2097152)
+ => ()
+  let buffer = make(<buffer>, size: block-size);
+  let address 
+    = primitive-wrap-machine-word
+        (primitive-cast-pointer-as-raw
+          (primitive-string-as-raw(buffer)));
+
+  block (finished?)
```
cgay, line 135: It's a little unusual for the exit-block function to be named as a predicate (with "?").  I'd just call it "finished" or "done".  There's no need to pass any arguments (`#t`) to it since unix-copy-file returns no values.

sources/system/file-system/unix-interface.dylan
```diff
@@ -106,6 +123,32 @@ define function unix-delete-file (path :: <byte-string>) => (ok :: <boolean>)
...
+    while (#t)
+      let bytes-read = unix-raw-read(source-fd, address, block-size);
+      select (bytes-read)
+         0 => finished?(#t);
+        -1 => unix-file-error("read", "unix-copy-file");
```
cgay, line 140: I think this and the call to `unix-file-error` below should display the source and destination file names.

`unix-file-error("read", "copy-file %s to %s", source, destination)`

I guess you'll need to pass in the source and destination locators to `unix-copy-file` so they can be displayed in the error message.

sources/system/unix-portability.c
```diff
@@ -2,6 +2,8 @@
 #define _DARWIN_C_SOURCE
 #endif
 
+#define _GNU_SOURCE
```
cgay, line 5: Just curious...what does this do?

cgay (changes requested):
One of us is going to have a fun merge. Sorry about that.

sources/system/unix-portability.c
fraya: I need it to use 'copy_file_range' that is not included in the POSIX version of `<unistd.h>`. Probably I should use something like `#ifdef __LINUX__ #define _GNU_SOURCE #endif`.

Macro: _GNU_SOURCE

If you define this macro, everything is included: ISO C89, ISO C99, POSIX.1, POSIX.2, BSD, SVID, X/Open, LFS, and GNU extensions. In the cases where POSIX.1 conflicts with BSD, the POSIX definitions take precedence.

sources/system/file-system/unix-file-system.dylan
```diff
@@ -195,13 +195,35 @@ define function %delete-file
...
+          end;
+          file-descriptor
+        end,
+        method open-source-file(filename)
+          let flags-mode   = $O_RDONLY;
+          let flags-create = $O_RDONLY;
+          open-file(filename, flags-mode, flags-create);
+        end,
+        method open-destination-file(filename)
+          let flags-mode   = logior($O_WRONLY, $O_CREAT, $O_TRUNC);
```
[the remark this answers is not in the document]
fraya: This is my proposal:

```dylan
 local method open-file(locator, flags-mode, flags-create)
          let file-descriptor
            = unix-open(as(<string>, locator), flags-mode, flags-create);
          if (file-descriptor < 0)
            select (unix-errno())
              $e_access =>
                error(make(<invalid-file-permissions-error>,
                           locator: as(<posix-file-locator>, locator)));
              $e_exists =>
                error(make(<file-exists-error>,
                           locator: as(<posix-file-locator>, locator)));
              otherwise =>
                unix-file-error("open", "file %s", as(<string>, locator));
            end select;
          end if;
          file-descriptor
        end method,
        method open-source-file(filename)
          let flags-mode   = $o_rdonly;
          let flags-create = 0;
          open-file(filename, flags-mode, flags-create);
        end method,
        method open-destination-file(filename, if-exists)
          let flags-mode 
            = if (if-exists = #"signal")
                logior($o_wronly, $o_creat, $o_trunc, $o_excl)
              else 
                logior($o_wronly, $o_creat, $o_trunc)
              end;
          let flags-create 
            = logior($s_irusr, $s_iwusr, $s_irgrp, $s_iwgrp, $s_iroth, $s_iwoth);
          open-file(filename, flags-mode, flags-create);
        end method;
```

sources/system/file-system/unix-file-system.dylan
```diff
@@ -214,14 +236,41 @@ define function %copy-file
...
+    if (failed?)
+      unix-copy-file(source-fd, destination-fd);
+    end;
+
+  cleanup
+    unless (source-fd < 0)
+      unix-close(source-fd)
+    end;
+    unless (destination-fd < 0)
+      unix-close(destination-fd)
```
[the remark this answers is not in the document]
fraya:
```dylan
    unless (source-fd < 0)
      if (unix-close(source-fd) = -1)
        unix-file-error("close", "close file %s", source)
      end
    end;
    unless (destination-fd < 0)
      if (unix-close(destination-fd) = -1)
        unix-file-error("close", "close file %s", destination)
      end
```

sources/system/file-system/unix-file-system.dylan
```diff
@@ -214,14 +236,41 @@ define function %copy-file
                format-arguments: list(as(<string>, source),
                                       as(<string>, destination))))
```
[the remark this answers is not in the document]
fraya: Done in a different way

sources/system/file-system/unix-interface.dylan
fraya: I would prefer to not change the interface passing source and destination because all the functions in this file has the same level of abstraction, using file descriptors. What I propose is return a status code: 0 for success, -1 for error reading and -2 for error writing. The caller can call 'unix-file-error' with the file names.

Here is the new code:
```dylan
define function unix-copy-file
    (source-fd :: <integer>, destination-fd :: <integer>, #key block-size = 2097152)
 => (status :: <integer>)
  let buffer = make(<buffer>, size: block-size);
  let address 
    = primitive-wrap-machine-word
        (primitive-cast-pointer-as-raw
          (primitive-string-as-raw(buffer)));

  block (status)
    while (#t)
      let bytes-read = unix-raw-read(source-fd, address, block-size);
      select (bytes-read)
         0 => status(0);  // finished read
        -1 => status(-1); // error reading
      otherwise => 
        let bytes-written 
          = unix-raw-write(destination-fd, address, bytes-read);
        unless (bytes-written > 0)
          status(-2);     // error writing
        end;
      end select;
    end while;
  end block;
end function unix-copy-file;
```
The caller:
```dylan
   // fallback copy in case of error
    if (failed?)
      let status = unix-copy-file(source-fd, destination-fd);
      unless (status = 0) 
        unix-file-error(if (status = -1) "read" else "write" end, 
                        "copy from file %s to %s",
                        source, destination)
      end;
    end;
```

sources/system/unix-portability.c
fraya: This is one idea:

```C
#if defined(OPEN_DYLAN_PLATFORM_LINUX)
#define _GNU_SOURCE
#endif
```

```C
/*
  Copy from 'in_fd' file (with 'in_size' bytes) to 'out_fd'.
  Return 0 on success and -1 if there is an error.
*/
int system_copy_file_range(int in_fd, int out_fd, off_t in_size);

#ifdef __APPLE__
#include <copyfile.h>

int system_copy_file_range(int in_fd, int out_fd, off_t in_size)
{
  return fcopyfile(in_fd, out_fd, NULL, COPYFILE_ALL);
}

#endif

// FreeBSD needs <sys/types.h> for 'copy_file_range'
#if defined(OPEN_DYLAN_PLATFORM_FREEBSD)
#include <sys/types.h>
#endif

#if defined(OPEN_DYLAN_PLATFORM_LINUX) || defined(OPEN_DYLAN_PLATFORM_FREEBSD)

int system_copy_file_range(int in_fd, int out_fd, off_t in_size)
{
  // offset: read from file adjusted by bytes copied
  off_t *_Nullable offset = NULL;
  // flags: used for future expansion, currently must be 0
  unsigned int flags = 0;
  // bytes copied
  off_t copied = -1;

  do {
    copied = copy_file_range(in_fd, offset, out_fd, offset, in_size, flags);
    if (copied == -1) return -1;
    in_size -= copied;
  } while (in_size > 0 && copied > 0);

  return 0;
}

#endif
``

sources/system/file-system/unix-interface.dylan
fraya: Done in next comment.

PR #1752, opened 2025-09-24
Improve copy-down-method and io performance

### LINKED ISSUE #1743: Why does map-as(<vector>, identity, {<set>}) return a <deque>?
https://play.opendylan.org/shared/e7357f76ff6d0eb3

```dylan
let collection
  = begin
      let s = make(<set>); add!(s, 1); s
    end;
format-out("%=", map-as(<vector>, identity, collection));
// => {deque of 1}
```

`<deque>` is not a subclass of `<vector>`.

Potentially related to #1097 ?

### LINKED ISSUE #1749: Lambda lifting loses types
The lambda lifting done by the `analyze-environments` pass (`flow-graph/closure.dylan`) adds new arguments to lambdas so that they can be lifted to top-level. Unfortunately the type used for these arguments in the modified lambda's signature is always `<object>`, potentially losing type context that might have been available in the outer environment.

housel:
Fixes #1749
Fixes #1743


cgay (commented):
I think the CI is failing because `define copy-down-method` is defined in the compiler and so the new `specializing` clause doesn't exist in the `install-opendylan` GitHub Action's compiler.

Also I think there may be some Tab chars again because some indentation looks out of whack.
Screenshot 2025-09-24 at 5 37 07 PM

PR #1788, opened 2026-02-01
dfmc-debug-back-end: Fix print-object(<simple-error>, <stream>)

cgay:
Because format-string: and format-arguments: are not required init keywords for `<simple-condition>` they may take on the default values. Some subclasses (notably subclasses of `<file-system-error>`) leave them empty and use their own condition-to-string method, so we fall back to that here.

The problem showed up when dswank got a `<file-does-not-exist-error>` and Emacs received the error message "Error: \n" rather than "File .../build.jam does not exist".

### DIFF
--- a/sources/dfmc/debug-back-end/print-condition.dylan
+++ b/sources/dfmc/debug-back-end/print-condition.dylan
@@ -107,8 +107,17 @@ end method print-object;
 
 define compiler-sideways method print-object
     (condition :: <simple-error>, stream :: <stream>) => ()
-  let body = apply(format-to-ppml, condition.condition-format-string,
-                                   condition.condition-format-arguments);
+  let format-string = condition.condition-format-string;
+  let format-arguments = condition.condition-format-arguments;
+  if (format-string.empty?)
+    // Because format-string: and format-arguments: are not required init keywords for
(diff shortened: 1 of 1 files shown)


sources/dfmc/debug-back-end/print-condition.dylan
```diff
@@ -107,8 +107,17 @@ end method print-object;
...
-  let body = apply(format-to-ppml, condition.condition-format-string,
-                                   condition.condition-format-arguments);
+  let format-string = condition.condition-format-string;
+  let format-arguments = condition.condition-format-arguments;
+  if (format-string.empty?)
+    // Because format-string: and format-arguments: are not required init keywords for
+    // <simple-condition> they may take on the default values. Some subclasses (notably
+    // subclasses of <file-system-error>) leave them empty and use their own
+    // condition-to-string method, so we fall back to that here.
+    format-string := condition-to-string(condition);
```
housel: Suggest using "%s" as the format-string and vector(condition-to-string(condition)) as the format-arguments

cgay: Thanks, done.

PR #1807, opened 2026-05-07
Update the LLVM back-end calling conventions to not use varargs

housel:
These commits change the LLVM bas-end to use a set of calling conventions that is compatible with a wider variety of platform ABIs. This change improves performance in certain situations (such as dispatching methods with keyword arguments).

### DIFF
--- a/sources/dfmc/llvm-back-end/llvm-emit-computation.dylan
+++ b/sources/dfmc/llvm-back-end/llvm-emit-computation.dylan
@@ -718,6 +718,15 @@ define method emit-computation
   end if;
 end method;
 
+// Calling conventions:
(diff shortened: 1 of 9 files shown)


sources/dfmc/llvm-back-end/llvm-emit-computation.dylan
cgay, line 1: This file needs untabify.

sources/dfmc/llvm-back-end/llvm-entry-points.dylan
cgay, line 1: Needs untabify

cgay (commented):
Just noticed that a couple of files have tabs in them so the indentation looks off.  I didn't check everything.

PR #1808, opened 2026-05-09
Add a native port for aarch64-darwin

housel:
With these changes it is no longer necessary to target x86_64 on Apple Silicon machines.

Closes #1562


sources/dfmc/llvm-back-end/llvm-targets.dylan
```diff
@@ -228,6 +228,33 @@ define method llvm-back-end-data-layout
...
+define method llvm-back-end-data-layout
+    (back-end :: <llvm-aarch64-darwin-back-end>) => (layout :: <string>);
+  "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-n32:64-S128-Fn32"
+end method;
+
+define method llvm-back-end-calling-convention-fast
+    (back-end :: <llvm-aarch64-darwin-back-end>)
+ => (calling-convention :: <integer>);
+  // We're forced to use the ordinary C calling convention for IEPs
+  // and primitives on AArch64 Darwin as well.
```
cgay: For the same reason as on line 185?  Maybe copy that comment here too.

housel: Not the same reason, because in fact it's not true (determined experimentally).

PR #1822, opened 2026-06-23
<table> implementation improvements

housel:


sources/dylan/table.dylan
```diff
@@ -74,10 +73,11 @@ define function search-for-entry-count
           // smallest good entry count that is not smaller than the
           // needed number of entries.
           if (lower = upper)
-            vector-element($entry-counts, lower);
+            primitive-the(<integer>, vector-element($entry-counts, lower))
```
cgay, line 76: Just curious, couldn't this also be achieved by making `$entry-counts` be a `limited(<vector>, of: <integer>)`?

housel: > Just curious, couldn't this also be achieved by making `$entry-counts` be a `limited(<vector>, of: <integer>)`?

Not quite as easily, you'd have to initialize it with `as(limited(<vector>, of: <integer>), #[...])` which would have an initialization-time cost. I don't think the compiler can do the conversion at compile time (yet).

PR #1823, opened 2026-06-26
Make the base-address project option obsolete

housel:
Since Windows Vista and Windows Server 2008, DLLs are loaded at random addresses as part of Address Space Layout Randomization. Windows XP, the last version that respected DLL base addresses specified in the DLL executable, ended support in April 2014. It is no longer desirable to specify base addresses for Dylan library projects.

### DIFF
--- a/documentation/source/library-reference/lid.rst
+++ b/documentation/source/library-reference/lid.rst
@@ -415,25 +415,21 @@ LID keyword
 
     Base-Address: *address*
 
+.. deprecated:: 2026.2
+   Because Windows versions starting with Windows Vista and
+   Windows Server 2008 implement Address Space Layout Randomization for
+   improved security, providing an initial base address is no longer
+   useful.
+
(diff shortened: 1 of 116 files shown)


documentation/source/library-reference/lid.rst
```diff
@@ -418,22 +418,17 @@ LID keyword
 .. note:: This keyword is only used on Windows and is ignored on other
    platforms.
 
-Specifies the base address of the DLL built from this Dylan library. The
-*address* must be a hexadecimal value. For convenience, you can use
-either Dylan (``#xNNNNNNNN``) or C (``0xNNNNNNNN``) notations when
-specifying the address.
+This keyword specifies the base address of the Windows DLL built from
+this Dylan library. Since Windows versions since Windows Vista and
```
cgay: "Since Windows versions since" reads a bit better as "Because Windows versions since".

You could also change the `.. note::` above to `.. version-deprecated:: 2026.2`

PR #1830, opened 2026-07-10
Post-release tasks

cgay:


documentation/source/hacker-guide/topics/making-a-release.rst
```diff
@@ -81,7 +81,7 @@ now here is a manual check-list.
      $ git clone --recursive https://github.com/dylan-lang/opendylan
      $ cd opendylan
      $ git co v2019.1.0      # the tag you created above
-     $ ./build/unix/release-with-batteries.sh
+     $ ./build/unix/container-release.sh
```
housel, line 84: This only applies to Linux; I was going to make a change that called out both Linux and non-Linux cases (but forgot)

housel (approved):

PR #1835, opened 2026-08-14
doc: various minor changes

cgay:


documentation/source/conf.py
```diff
@@ -33,6 +33,9 @@
 show_authors = True
 templates_path = ['_templates']
 
+# Work around https://github.com/sphinx-doc/sphinx/issues/13904 Specifically, I (cgay)
```
housel, line 36: You did figure it out back in April; the problem is at idl-app.rst line 1615.

housel (approved):

documentation/source/conf.py
cgay: Thanks, I fixed that one, but sadly that's only one of the problems.

